PR #30, opened 2016-10-11
Ak/handle nils in portfolio balance presenters

mitochondrion:


guillaumedebavelaere:
@mitochondrion I finished refactoring, see the 2 last commits

guillaumedebavelaere:
@mitochondrion I finished refactoring, see the 
3 last commits

TradeItIosTicketSDK2/TradeItPortfolioBalanceFXPresenter.swift
```diff
@@ -1,30 +1,59 @@
...
+    
+    func getFormattedTotalValueWithPercentage() -> String{
+        guard let totalValueBaseCurrency = fxBalance.totalValueBaseCurrency
+            else { return TradeItPresenter.MISSING_DATA_PLACEHOLDER}
+        
+        var formattedTotalValue = NumberFormatter.formatCurrency(totalValueBaseCurrency)
+
+        let totalReturn = fxBalance.unrealizedProfitAndLossBaseCurrency?.floatValue
+        if (totalReturn != nil && totalReturn != 0) {
+            let totalPercentReturn = totalReturn! / (totalValueBaseCurrency.floatValue - abs(totalReturn!))
```
jsom: We should avoid using `!` in places like this. You could maybe extract the percentage out to another function and use another guard:

```
func getTotalPercentage() -> String {
  guard let totalValueBaseCurrency = fxBalance.totalValueBaseCurrency?.floatValue,
    let totalReturn = fxBalance.unrealizedProfitAndLossBaseCurrency?.floatValue
    where totalReturn != 0
    else { return "" }

  let totalPercentReturn = totalReturn / (totalValueBaseCurrency - abs(totalReturn))
  return " (" + NumberFormatter.formatPercentage(totalPercentReturn) + ")"
}
```

jsom: Do we need a `+/-` sign in the percentage?

TradeItIosTicketSDK2/TradeItPortfolioPosition.swift
```diff
@@ -22,29 +22,36 @@ class TradeItPortfolioPosition : NSObject {
...
         self.tradeItMarketDataService.getQuoteData(tradeItQuoteRequest, withCompletionBlock: { (tradeItResult: TradeItResult!) -> Void in
             if let tradeItQuoteResult = tradeItResult as? TradeItQuotesResult {
-                let results = tradeItQuoteResult.quotes.filter { return $0.symbol == symbol}
-                if results.count > 0 {
-                    quote = results[0] as! TradeItQuote
-                    self.quote = quote
+                let results = tradeItQuoteResult.quotes?.filter { return $0.symbol == symbol }
+
+                if let results = results as? [TradeItQuote] where results.count > 0 {
+                    self.quote = results[0]
```
jsom: You can also use `self.quote = results.first` which won't explode if there is no element - it will just return `nil`.

TradeItIosTicketSDK2/TradeItPortfolioPositionPresenter.swift
```diff
@@ -1,63 +1,71 @@
...
+        return formatCurrency(low) + " - " + formatCurrency(high)
     }
+}
 
-    func getQuantity() -> Float {
-        return 0.0
+class TradeItPortfolioPositionPresenterFactory {
+    
+    static func forTradeItPortfolioPosition(tradeItPortfolioPosition: TradeItPortfolioPosition) -> TradeItPortfolioPositionPresenter {
+        if tradeItPortfolioPosition.position != nil {
```
jsom: Can we use `if let` here?

TradeItIosTicketSDK2/TradeItPortfolioBalanceEquityPresenter.swift
```diff
@@ -1,30 +1,75 @@
...
 
-class TradeItPortfolioBalanceEquityPresenter {
+class TradeItPortfolioBalanceEquityPresenter: TradeItPortfolioBalancePresenter {
 
-    let balance: TradeItAccountOverview
+    var balance: TradeItAccountOverview = TradeItAccountOverview()
     
     init(_ tradeItLinkedBrokerAccount: TradeItLinkedBrokerAccount) {
-        self.balance = tradeItLinkedBrokerAccount.balance
+        if let balance = tradeItLinkedBrokerAccount.balance {
```
mitochondrion: Good point.  Updated.

TradeItIosTicketSDK2/TradeItPortfolioBalanceFXPresenter.swift
guillaumedebavelaere: What happens if total return is nil but not totalValueBaseCurrency ? Because in that case we want to return totalValueBaseCurrency and not ""

TradeItIosTicketSDK2/TradeItPortfolioPositionPresenter.swift
guillaumedebavelaere: We could but it adds a warning because we don't use the variable

jsom: That's true -- is it possible for a `PortfolioPosition` to have both a `position` and an `fxPosition`?

TradeItIosTicketSDK2/TradeItPortfolioBalanceFXPresenter.swift
jsom: This function only calculates the percentage, so it would be called by the other function you wrote.

mitochondrion: Yup.  Fixed.

TradeItIosTicketSDK2/TradeItPortfolioPosition.swift
mitochondrion: Done.

PR #35, opened 2016-10-14
gd/exit_mechanism

guillaumedebavelaere:


TradeItIosTicketSDK2/UIViewController+TradeIt.swift
```diff
@@ -0,0 +1,33 @@
+import UIKit
+
+extension UIViewController {
```
jsom: We should not extend UIKit classes. We should create a protocol like `TradeItViewController` that we extend with these methods instead.

PR #39, opened 2016-10-19
Fix UI test crashing for missing dynamic linked lib

jsom:
The solution was to add a Build Phase to the ExampleApp to copy the TradeItIosTicketSDK2.framework file over.

The actual error messages were being swallowed by a filter applied to the logs that stripped out networking messages.


Podfile
```diff
@@ -13,19 +13,19 @@ def app_pods
   pod 'PromiseKit', '3.4'
 end
 
-target 'TradeItIosTicketSDK2Tests' do
-    testing_pods
-    app_pods
-end
+# target 'TradeItIosTicketSDK2Tests' do
```
mitochondrion: Uncomment and reinstate unit tests

ExampleApp/Info.plist
```diff
@@ -30,22 +28,11 @@
 	<array>
 		<string>armv7</string>
 	</array>
-	<key>NSAppTransportSecurity</key>
```
mitochondrion: d'oh.  we need this to hit the server on localhost

PR #40, opened 2016-10-19
gd/avg cost

guillaumedebavelaere:


TradeItIosTicketSDK2/TradeItPortfolioEquityPositionPresenter.swift
```diff
@@ -16,11 +16,8 @@ class TradeItPortfolioEquityPositionPresenter: TradeItPortfolioPositionPresenter
         return symbol
     }
 
-    func getQuantity() -> Float? {
-        guard let quantity = self.position?.quantity
-            else { return 0 }
-        
-        return quantity.floatValue
+    func getQuantity() -> NSNumber? {
+        return self.position?.quantity
```
jsom: Much cleaner! Do we still want to default it to 0 though?

TradeItIosTicketSDK2/TradeItPortfolioEquityPositionPresenter.swift
```diff
@@ -53,11 +50,11 @@ class TradeItPortfolioEquityPositionPresenter: TradeItPortfolioPositionPresenter
         return NumberFormatter.formatPercentage(totalGainLossPercentage.floatValue);
     }
 
-    func getCostBasis() -> String {
-        guard let costBasis = self.position?.costbasis
+    func getAvgCostBasis() -> String {
```
jsom: Is there a different meaning in `getAvgCostBasis` vs `getAvgCost`? If there is no difference, then we should stick with something consistent like `getAvgCost`.

TradeItIosTicketSDK2/TradeItPortfolioFxPositionPresenter.swift
```diff
@@ -16,8 +16,8 @@ class TradeItPortfolioFxPositionPresenter: TradeItPortfolioPositionPresenter {
         return symbol
     }
     
-    func getQuantity() -> Float? {
-        return self.fxPosition.quantity as? Float
+    func getQuantity() -> NSNumber? {
+        return self.fxPosition.quantity
```
jsom: How come `fxPosition` isn't optional here?

TradeItIosTicketSDK2/TradeItPortfolioEquityPositionPresenter.swift
guillaumedebavelaere: I removed the default to 0, because somewhere else we use the getQuantity method and we want to return N/A if it is not available

TradeItIosTicketSDK2/TradeItPortfolioFxPositionPresenter.swift
guillaumedebavelaere: We initialize the fxPosition at the beginning:

`var fxPosition: TradeItFxPosition = TradeItFxPosition()`

PR #41, opened 2016-10-19
gd/multiple choice security

guillaumedebavelaere:


TradeItIosTicketSDK2/TradeItAlertProvider.swift
```diff
@@ -34,28 +34,41 @@ class TradeItAlertProvider {
...
             title: "Cancel",
             style: UIAlertActionStyle.Default) { (result : UIAlertAction) -> Void in
                 onCancelSecurityQuestion()
+        }
+        
+        alertController.addAction(cancelAction)
+        
+        if multipleOptions.count > 0 {
+            for option in multipleOptions {
+                let optionAction = UIAlertAction(title: option, style: .Default, handler: { (action) -> Void in
```
jsom: This is great as is, but just a small suggestion: you don't always need to provide the argument types or return types in these closures so `{ action in` works just fine and is still type checked because the `UIAlertAction` has already defined what type signature the handler should be.

jsom:
Looks good to me. Merge when ready.

PR #43, opened 2016-10-20
refactors to use switch instead of if to check types

guillaumedebavelaere:


guillaumedebavelaere:
There is also the marketDataService, but I don't know if we can handle with the switch case several "let" in the same "if" :

```
marketDataService.getQuoteData(quotesRequest, withCompletionBlock: { (tradeItResult: TradeItResult!) in
            if let quotesResult = tradeItResult as? TradeItQuotesResult,
                let quote = quotesResult.quotes?.first as? TradeItQuote {
                onSuccess(quote)
            } else if let errorResult = tradeItResult as? TradeItErrorResult {
                onFailure(errorResult)
            } else {
                onFailure(TradeItErrorResult.tradeErrorWithSystemMessage("Error loading market data quote. Please try again later."))
            }
        })
```

jsom:
LGTM and I think it's fine to leave the Market Data Service as is.

PR #45, opened 2016-10-20
Gd/authenticate for trading

guillaumedebavelaere:


TradeItIosTicketSDK2/TradeItErrorResult+ErrorCodeHelpers.swift
```diff
@@ -26,10 +26,6 @@ extension TradeItErrorResult {
     }
 
     func requiresAuthentication() -> Bool {
-        guard let integerCode = self.code?.integerValue
-            , let errorCode = TradeItErrorCode(rawValue: integerCode)
-            else { return false }
-
-        return errorCode == TradeItErrorCode.SESSION_ERROR || requiresRelink()
+        return !requiresRelink()
```
guillaumedebavelaere: We need to authenticate on all errors except the one which need to relink the account

TradeItIosTicketSDK2/TradeItTradingTicketViewController.swift
```diff
@@ -92,34 +92,36 @@ class TradeItTradingTicketViewController: TradeItViewController, TradeItSymbolSe
...
-            onSuccess: {
-                self.ezLoadingActivityManager.updateText(text: "Previewing Order")
-                self.order.preview(onSuccess: { previewOrder, placeOrderCallback in
-                    self.ezLoadingActivityManager.hide()
-                    self.delegate?.tradeItTradingTicketViewController(self,
-                        previewOrder: previewOrder,
-                        placeOrderCallback: placeOrderCallback)
-                    
-                    }, onFailure: { errorResult in
+        if let linkedBroker = self.order.linkedBrokerAccount?.linkedBroker {
```
jsom: Using a guard here would reduce the nesting in this function. Up to you though.

TradeItIosTicketSDK2/TradeItTradingTicketViewController.swift
```diff
@@ -91,18 +91,37 @@ class TradeItTradingTicketViewController: TradeItViewController, TradeItSymbolSe
...
+                onSecurityQuestion: { (securityQuestion: TradeItSecurityQuestionResult, answerSecurityQuestion: (String) -> Void, cancelSecurityQuestion: () -> Void) -> Void in
+                    self.ezLoadingActivityManager.hide()
+                    self.alertManager.show(
+                        securityQuestion: securityQuestion,
+                        onViewController: self,
+                        onAnswerSecurityQuestion: answerSecurityQuestion,
+                        onCancelSecurityQuestion: cancelSecurityQuestion
+                    )
+                },
+                onFailure: { (tradeItErrorResult: TradeItErrorResult) -> Void in
```
jsom: We don't need to prefix the variables with `tradeIt` everywhere or give the type signatures/return value for the function

TradeItIosTicketSDK2/TradeItTradingTicketViewController.swift
```diff
@@ -91,18 +91,37 @@ class TradeItTradingTicketViewController: TradeItViewController, TradeItSymbolSe
...
-            self.delegate?.tradeItTradingTicketViewController(self,
-                previewOrder: previewOrder,
-                placeOrderCallback: placeOrderCallback)
-            
-        }, onFailure: { errorResult in
-            self.ezLoadingActivityManager.hide()
-            self.alertManager.show(tradeItErrorResult: errorResult, onViewController: self, withLinkedBroker: self.order.linkedBrokerAccount!.linkedBroker, onFinished: {})
-        })
+        self.ezLoadingActivityManager.show(text: "Authenticating", disableUI: true)
+        if let linkedBroker = self.order.linkedBrokerAccount?.linkedBroker {
```
mitochondrion: Should be guard:
```
guard let linkedBroker = self.order.linkedBrokerAccount?.linkedBroker else { return }
```

jsom:
So `authenticateIfNeeded` will only work if there was some previous request that failed I think. What about if *this* request fails? Isn't that when we want to authenticate and then retry the operation?

TradeItIosTicketSDK2/TradeItTradingTicketViewController.swift
```diff
@@ -91,18 +91,37 @@ class TradeItTradingTicketViewController: TradeItViewController, TradeItSymbolSe
...
+                        self.delegate?.tradeItTradingTicketViewController(self,
+                            previewOrder: previewOrder,
+                            placeOrderCallback: placeOrderCallback)
+                        
+                        }, onFailure: { errorResult in
+                            self.ezLoadingActivityManager.hide()
+                            self.alertManager.show(tradeItErrorResult: errorResult, onViewController: self, withLinkedBroker: linkedBroker, onFinished: {})
+                    })
+                },
+                onSecurityQuestion: { (securityQuestion: TradeItSecurityQuestionResult, answerSecurityQuestion: (String) -> Void, cancelSecurityQuestion: () -> Void) -> Void in
```
mitochondrion: E.g.:
```
onSecurityQuestion: { (securityQuestion, answerSecurityQuestion, cancelSecurityQuestion) -> Void in
```

PR #46, opened 2016-10-21
Fix error messages

jsom:
Using `systemMessage` on a `TradeItErrorResult` was useless in the cases I was using them. I created a new convenience initializer for errors to be displayed to the user and converted my system message errors to use it.

### DIFF
--- a/TradeItIosTicketSDK2/TradeItAlertManager.swift
+++ b/TradeItIosTicketSDK2/TradeItAlertManager.swift
@@ -4,67 +4,65 @@ import UIKit
     var linkedBrokerManager = TradeItLauncher.linkedBrokerManager
     var linkBrokerUIFlow = TradeItLinkBrokerUIFlow(linkedBrokerManager: TradeItLauncher.linkedBrokerManager)
 
-    func showGenericError(tradeItErrorResult tradeItErrorResult: TradeItErrorResult,
-                            onViewController viewController: UIViewController,
-                                             onFinished: () -> Void = {}) {
-        let alertTitle = tradeItErrorResult.shortMessage ?? ""
-        let messages = (tradeItErrorResult.longMessages as? [String]) ?? []
-        let alertMessage = messages.joinWithSeparator(" ")
-        let alertActionTitle = "OK"
+    public func showError(error: TradeItErrorResult,
+                          onViewController viewController: UIViewController,
+                          onFinished: () -> Void = {}) {
+        let title = error.shortMessage ?? ""
+        let messages = (error.longMessages as? [String]) ?? []
+        let message = messages.joinWithSeparator(". ")
+        let actionTitle = "OK"
 
-        self.showOn(viewController: viewController,
-                    withAlertTitle: alertTitle,
-                    withAlertMessage: alertMessage,
-                    withAlertActionTitle: alertActionTitle,
+        self.showAlert(onViewController: viewController,
+                              withTitle: title,
+                            withMessage: message,
+                        withActionTitle: actionTitle,
                     onAlertActionTapped: onFinished)
     }
 
-    public func show(tradeItErrorResult tradeItErrorResult: TradeItErrorResult,
-                onViewController viewController: UIViewController,
-                withLinkedBroker linkedBroker: TradeItLinkedBroker,
-                                 onFinished : () -> Void) {
+    public func showRelinkError(error: TradeItErrorResult,
+                                withLinkedBroker linkedBroker: TradeItLinkedBroker,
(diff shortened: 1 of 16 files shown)


TradeItIosTicketSDK2/TradeItTradingTicketViewController.swift
```diff
@@ -99,9 +99,18 @@ class TradeItTradingTicketViewController: TradeItViewController, TradeItSymbolSe
                 previewOrder: previewOrder,
                 placeOrderCallback: placeOrderCallback)
             
-        }, onFailure: { errorResult in
+        }, onFailure: { error in
+            guard let linkedBroker = self.order.linkedBrokerAccount?.linkedBroker else {
+                return assertionFailure("TradeItIosTicketSDK ERROR: TradeItTradingTicketViewController loaded without setting linkedBrokerAccount on order.")
```
jsom: How do we want to handle this in release mode?

TradeItIosTicketSDK2/TradeItLoginViewController.swift
```diff
@@ -59,32 +59,26 @@ class TradeItLoginViewController: KeyboardViewController {
...
+            self.linkedBrokerManager.relinkBroker(
+                linkedBrokerToRelink,
+                authInfo: tradeItAuthenticationInfo,
+                onSuccess: self.authenticateBroker,
+                onFailure: { error in
+                  self.activityIndicator.stopAnimating()
+                  self.enableLinkButton()
+                  self.alertManager.showError(error, onViewController: self)
+                }
+            )
```
jsom: I think formatting things this way is much clearer.

TradeItIosTicketSDK2/TradeItLinkedBrokerAccount.swift
```diff
@@ -43,7 +43,7 @@ public class TradeItLinkedBrokerAccount: NSObject {
                 self.linkedBroker.error = errorResult
                 onFailure(errorResult)
             default:
-                onFailure(TradeItErrorResult.tradeErrorWithSystemMessage("Unknown error getting balances"))
+                onFailure(TradeItErrorResult(title: "Balances failed"))
```
mitochondrion: Change to: "Failed to retrieve account balances"

TradeItIosTicketSDK2/TradeItLinkedBrokerAccount.swift
```diff
@@ -73,7 +73,7 @@ public class TradeItLinkedBrokerAccount: NSObject {
                 self.linkedBroker.error = errorResult
                 onFailure(errorResult)
             default:
-                onFailure(TradeItErrorResult.tradeErrorWithSystemMessage("Unknown error getting positions"))
+                onFailure(TradeItErrorResult(title: "Positions failed"))
```
mitochondrion: Change to: "Failed to retrieve account positions"

TradeItIosTicketSDK2/TradeItTradingTicketViewController.swift
mitochondrion: This should probably push the account selection flow instead of failing.

TradeItIosTicketSDK2/TradeItAlertManager.swift
```diff
@@ -4,67 +4,65 @@ import UIKit
...
+                            withMessage: message,
+                        withActionTitle: actionTitle,
                     onAlertActionTapped: onFinished)
     }
 
-    public func show(tradeItErrorResult tradeItErrorResult: TradeItErrorResult,
-                onViewController viewController: UIViewController,
-                withLinkedBroker linkedBroker: TradeItLinkedBroker,
-                                 onFinished : () -> Void) {
+    public func showRelinkError(error: TradeItErrorResult,
```
jsom: This feels more like a "recover from error if possible" helper. The user would have to implement the logic for checking the type of error otherwise which feels unnecessary to me.

TradeItIosTicketSDK2/TradeItTradingTicketViewController.swift
jsom: Made a Tracker story.

TradeItIosTicketSDK2/TradeItAlertManager.swift
jsom: This will be a larger refactor. I've put a story in it. We are using it in lots of places which may get other errors like connection failures as well. Now all of those places will need to determine the error type and call `promptUserToRelink`.

PR #47, opened 2016-10-24
Gd/fix important unit tests

guillaumedebavelaere:


TradeItIosTicketSDK2/TradeItLinkedBrokerManager.swift
```diff
@@ -5,15 +5,23 @@ import PromiseKit
...
-        tradeItConnector = TradeItConnector(apiKey: apiKey, environment: environment, version: TradeItEmsApiVersion_2)!
-        tradeItSessionProvider = TradeItSessionProvider()
+        self.tradeItConnector = TradeItConnector(apiKey: apiKey, environment: environment, version: TradeItEmsApiVersion_2)!
+        self.tradeItSessionProvider = TradeItSessionProvider()
 
         super.init()
         self.loadLinkedBrokersFromKeychain()
     }
     
+    init(tradeItConnector: TradeItConnector) {
```
guillaumedebavelaere: Don't know if there is another way, but we need to inject a FakeTradeItConnector in our unit test

TradeItIosTicketSDK2Tests/TradeItLinkedBrokerManagerSpec.swift
```diff
@@ -596,10 +595,10 @@ class TradeItLinkedBrokerManagerSpec: QuickSpec {
                 expect(linkedBroker1.calls.forMethod("refreshAccountBalances(onFinished:)").count).to(equal(1))
                 expect(linkedBroker2.calls.forMethod("refreshAccountBalances(onFinished:)").count).to(equal(1))
             }
-
-            it("doesn't refresh the unauthenticated linkedBroker") {
-                expect(linkedBroker3.calls.forMethod("refreshAccountBalances(onFinished:)").count).to(equal(0))
-            }
+// Do we still want this behavior ?
```
guillaumedebavelaere: Before we were expecting to not refreshBalances on linkedBrokers that weren't authenticated. What behavior do we want ?

PR #51, opened 2016-10-25
gd/touch_id

guillaumedebavelaere:


TradeItIosTicketSDK2/TradeItDeviceManager.swift
```diff
@@ -0,0 +1,69 @@
...
+                                    onSuccess()
+                                }
+                                else {
+                                    guard let error = evaluationError else {
+                                        print("authenticateUserWithTouchId failed")
+                                        onFailure()
+                                        return
+                                    }
+                                    
+                                    switch error.code {
```
guillaumedebavelaere: Not sure how do we want to handle these errors. In the old SDK if the user cancel the authentication, we take him to the link broker screen, but it doesn't make sense to me, we should just deny access to him for launching portfolio or trading?

TradeItIosTicketSDK2/TradeItDeviceManager.swift
```diff
@@ -0,0 +1,69 @@
...
+
+class TradeItDeviceManager {
+    
+    func authenticateUserWithTouchId(onSuccess onSuccess: () -> Void, onFailure: () -> Void) {
+        var error: NSError?
+        let myLocalizedReasonString = "Authentication is required"
+        
+        let context = LAContext()
+        
+        guard context.canEvaluatePolicy(LAPolicy.DeviceOwnerAuthentication, error: &error) else {
```
guillaumedebavelaere: DeviceOwnerAuthentication policy will show the touchId if there is a fingerprint registered otherwise it will show the passcode

TradeItIosTicketSDK2/TradeItDeviceManager.swift
```diff
@@ -0,0 +1,69 @@
...
+            print("touch id not available on this device: \(error)")
+            onSuccess()
+            return
+        }
+        
+        context.evaluatePolicy(LAPolicy.DeviceOwnerAuthentication,
```
jsom: I would put all of this on a newline that is indented once.

TradeItIosTicketSDK2/TradeItDeviceManager.swift
```diff
@@ -0,0 +1,69 @@
...
+                                        onFailure()
+                                    case LAError.SystemCancel.rawValue:
+                                        print("Canceled by system (another application went to foreground)")
+                                        onFailure()
+                                    case LAError.PasscodeNotSet.rawValue:
+                                        print("Passcode not set")
+                                        onFailure()
+                                    case LAError.TouchIDNotAvailable.rawValue:
+                                        print("Touch id not available on the device")
+                                        onFailure()
```
jsom: If they have neither passcode or touch ID it looks like this will just fail? We probably just want to call `onSuccess` in that case.

TradeItIosTicketSDK2/TradeItDeviceManager.swift
jsom: Yeah, I agree. We should just leave them where they currently are before they hit Trade or whatever.

TradeItIosTicketSDK2/TradeItDeviceManager.swift
guillaumedebavelaere: I think it will be catch by the first guard, so we shouldn't have this case (I tested it on the test device). But correct, I am going to fix it.

PR #62, opened 2016-11-06
Add Carthage support

adamkaplan:
At the moment we are prevented from using CocoaPods with framework support due to a very old but extremely critical dependency. Luckily Carthage support was not difficult to add with 0 code changes.

I created a new target for Carthage – which is the only shared scheme – which will be picked up by Carthage users. It has no impact on CocoaPods users. All 3 of your upstream dependencies support Carthage natively.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -24,5 +24,6 @@ DerivedData
 *.xcuserstate
 
 Pods
+Carthage/
 
 .idea/
--- a/Cartfile
+++ b/Cartfile
@@ -1,0 +1,4 @@
+github "mxcl/PromiseKit" ~> 4.0
+github "jdg/MBProgressHUD" ~> 1.0.0
+github "radex/SwiftyUserDefaults" ~> 3.0.0
+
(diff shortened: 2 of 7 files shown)


jsom:
Thanks @adamkaplan! Will check this out today.

jsom:
Hey @adamkaplan re latest commit is it necessary to add another umbrella heaader or are you able to use this one? https://github.com/tradingticket/TradeItIosTicketSDK2/blob/develop/TradeItIosTicketSDK2/TradeItIosTicketSDK2.h

adamkaplan:
Yeah it absolutely is needed. I spent a full day on this! If the umbrella header name is *not* the same as `$(TARGET_NAME).h`, Xcode will `#import` it with an absolute path (starting at `/...`), which causes the Xcode on the other side – the client Xcode – to spit out a compile error for `Include of non-modular header inside framework module`.

This error can be bypassed, but then every client would need to disable that warning. As a last ditch effort, I copied the umbrella file and changed the name to match the target (or product), removed the manual override in the project file, and it just worked!

jsom:
@adamkaplan thanks for the detailed explanation. Confusing! I'm going to merge this now and then update the target with any new files added since you opened the PR.

PR #70, opened 2016-11-15
Add authentication observer for server syncing

jsom:


TradeItIosEmsApi/TradeItAuthLinkResult.h
```diff
@@ -11,8 +11,8 @@
 
 @interface TradeItAuthLinkResult : TradeItResult
 
-@property (nullable) NSString *userId;
-@property (nullable) NSString *userToken;
+@property (nonnull) NSString *userId;
+@property (nonnull) NSString *userToken;
```
jsom: I don't know if I can change these - the JSON lib we are using is a bit opaque. Were these set to `nullable` just for safety? Do we ever change our behaviour if they are `nil`? Not that I saw.

mitochondrion: I say leave them nullable.  The only way they would be null is a severe server bug, however, the app still shouldn't crash in that scenario, and the app can't prevent the server from returning null values in JSON.

PR #71, opened 2016-11-15
[#134277511] Add Help | Privacy | Terms

guillaumedebavelaere:


TradeItIosTicketSDK2/TradeItWebViewController.swift
```diff
@@ -0,0 +1,21 @@
...
+class TradeItWebViewController: UIViewController, UIWebViewDelegate {
+
+    @IBOutlet weak var webView: UIWebView!
+    var url = ""
+    var pageTitle = ""
+    
+    override func viewDidLoad() {
+        super.viewDidLoad()
+        self.navigationItem.title = "Loading...";
+        self.webView.loadRequest(URLRequest(url: URL (string: self.url)!))
```
jsom: We should not force unwrap here. We should `guard` instead and do nothing / dismiss the web view.

jsom:
@guillaumedebavelaere this is good to go! When creating new files now we need to make sure we add them to the Carthage build target. I've patched this one already.

PR #78, opened 2016-11-22
[#134610027] fixes when launching portfolio with selected account

guillaumedebavelaere:
There were 2 issues:
- We used Dictionary to store the data in userDefaults and Dictionary doesn't guaranty the order, so the accounts were stored in reverse order.
- In the portfolioviewcontroller we checked if the selected account (loaded from the cache) was in the account table (table of new TradeItLinkedBrokerAccount because we create new object when we authenticate), so the reference were not equal.

### DIFF
--- a/TradeItIosTicketSDK2/TradeItLinkedBrokerCache.swift
+++ b/TradeItIosTicketSDK2/TradeItLinkedBrokerCache.swift
@@ -1,14 +1,13 @@
 class TradeItLinkedBrokerCache {
     typealias UserId = String
-    typealias AccountNumber = String
     typealias SerializedLinkedBroker = [String: Any]
     typealias SerializedLinkedBrokers = [UserId: SerializedLinkedBroker]
     typealias SerializedLinkedBrokerAccount = [String: String]
-    typealias SerializedLinkedBrokerAccounts = [AccountNumber: SerializedLinkedBrokerAccount]
 
     private let ACCOUNTS_KEY = "ACCOUNTS"
     private let ACCOUNTS_LAST_UPDATED_KEY = "ACCOUNTS_LAST_UPDATED"
     private let ACCOUNT_NAME_KEY = "ACCOUNT_NAME"
+    private let ACCOUNT_NUMBER_KEY = "ACCOUNT_NUMBER"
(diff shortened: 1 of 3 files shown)


TradeItIosTicketSDK2/TradeItLinkedBrokerCache.swift
```diff
@@ -4,11 +4,11 @@ class TradeItLinkedBrokerCache {
     typealias SerializedLinkedBroker = [String: Any]
     typealias SerializedLinkedBrokers = [UserId: SerializedLinkedBroker]
     typealias SerializedLinkedBrokerAccount = [String: String]
-    typealias SerializedLinkedBrokerAccounts = [AccountNumber: SerializedLinkedBrokerAccount]
 
     private let ACCOUNTS_KEY = "ACCOUNTS"
     private let ACCOUNTS_LAST_UPDATED_KEY = "ACCOUNTS_LAST_UPDATED"
     private let ACCOUNT_NAME_KEY = "ACCOUNT_NAME"
+    private let ACCOUNT_NUMBER_KEY = "ACCOUNT_NUMBER"
```
guillaumedebavelaere: Moved the account number to the SerializedLinkedBrokerAccount map

TradeItIosTicketSDK2/TradeItPortfolioAccountsTableViewManager.swift
```diff
@@ -34,7 +34,7 @@ class TradeItPortfolioAccountsTableViewManager: NSObject, UITableViewDelegate, U
 
         if accounts.count > 0 {
             let selectedAccount = selectedAccount ?? self.accounts[0]
-            let selectedAccountIndex = self.accounts.index(of: selectedAccount) ?? 0
+            let selectedAccountIndex = self.accounts.index(where: {$0.accountNumber == selectedAccount.accountNumber && $0.brokerName == selectedAccount.brokerName}) ?? 0
```
guillaumedebavelaere: we are using the account number and the brokerName to get the selectedAccount index

jsom:
I think this makes sense. Maybe get @mitochondrion to have a think about it too before merging.

PR #80, opened 2016-11-30
Adds LinkedBrokerAccount.isEnabled to LinkedBrokerCache account caching

mitochondrion:
@jsom Plz review and add this to the new pod/carthage. Thx!


jsom:
@mitochondrion it's not super clear if it's covered in the test: but what will happen to existing serialized accounts that were added before this change?

PR #81, opened 2016-11-30
move didUnkink after removing the linkedBroker from the linkedBrokers…

guillaumedebavelaere:
…' list


jsom:
Thanks @guillaumedebavelaere!

PR #84, opened 2016-12-08
Fix order prepopulate

jsom:


TradeItIosTicketSDK2/TradeItTradingTicketViewController.swift
```diff
@@ -325,34 +351,24 @@ class TradeItTradingTicketViewController: TradeItViewController, TradeItSymbolSe
             , let linkedBrokerAccount = order.linkedBrokerAccount
             else { return }
 
-        linkedBrokerAccount.linkedBroker.authenticateIfNeeded(onSuccess: {
```
mitochondrion: Put this back?

ExampleApp/ExampleViewController.swift
```diff
@@ -59,6 +59,12 @@ class ExampleViewController: UIViewController, UITableViewDataSource, UITableVie
         case .launchTradingWithSymbol:
             let order = TradeItOrder()
             order.symbol = "CMG"
+            order.quantity = 10
```
mitochondrion: Add comment: `// Any order fields that are set will pre-populate the ticket.`

PR #85, opened 2016-12-09
Extract TradeItSDK class that holds launcher statically

jsom:


TradeItIosTicketSDK2/TradeItSDK.swift
```diff
@@ -0,0 +1,21 @@
+@objc public class TradeItSDK: NSObject {
+    private static var apiKey: String?
+    private static var environment: TradeitEmsEnvironments?
+
+    public static let launcher = TradeItLauncher()
+    private (set) public static var linkedBrokerManager: TradeItLinkedBrokerManager!
+    private (set) public static var marketDataService: TradeItMarketService!
+    private (set) public static var brokerCenterService: TradeItBrokerCenterService!
```
jsom: @mitochondrion any ideas for how to warn if they haven't called `configure` before? Computed properties won't allow you to throw an exception or anything like that.

TradeItIosTicketSDK2/TradeItSDK.swift
```diff
@@ -0,0 +1,21 @@
+@objc public class TradeItSDK: NSObject {
```
jsom: Can you actually check this branch out and run it. It compiles and runs, but I still get compilation ERRORS about `TradeItSDK` not found in various classes. It's really weird. I cleaned the build completely but I still see the errors in xcode.

TradeItIosTicketSDK2/TradeItSDK.swift
mitochondrion: Live example: http://swiftlang.ng.bluemix.net/#/repl/584b448223d4e234e48b7eae

```swift
class Foo
    private static var _thing: Int?
    public static var thing: Int {
        get {
            precondition(_thing != nil, "ERROR: Foo.thing referenced before calling Foo.config()!")
            return _thing!
        }
    }

    public static func config() {
        Foo._thing
    }
}
```

TradeItIosTicketSDK2/TradeItSelectBrokerViewController.swift
```diff
@@ -5,7 +5,7 @@ class TradeItSelectBrokerViewController: TradeItViewController, UITableViewDeleg
     @IBOutlet weak var brokerTable: UITableView!
     var delegate: TradeItSelectBrokerViewControllerDelegate?
     var alertManager = TradeItAlertManager()
-    var linkedBrokerManager: TradeItLinkedBrokerManager = TradeItLauncher.linkedBrokerManager
+    var linkedBrokerManager: TradeItLinkedBrokerManager = TradeItSDK.linkedBrokerManager
```
jsom: Yeah that's fine - I thought you preferred this way so was sticking with the pattern

TradeItIosTicketSDK2/TradeItSDK.swift
jsom: Hmm, interesting. I can't think of a better way of getting the desired effect. Let's do it.

TradeItIosTicketSDK2/TradeItSDK.swift
jsom: @mitochondrion pls do this. I know it makes no sense. Literally just check it out and run it and see if you get any compiler errors.

TradeItIosTicketSDK2/TradeItSDK.swift
```diff
@@ -0,0 +1,21 @@
...
+    private static var apiKey: String?
+    private static var environment: TradeitEmsEnvironments?
+
+    public static let launcher = TradeItLauncher()
+    private (set) public static var linkedBrokerManager: TradeItLinkedBrokerManager!
+    private (set) public static var marketDataService: TradeItMarketService!
+    private (set) public static var brokerCenterService: TradeItBrokerCenterService!
+
+    public static func configure(apiKey: String, environment: TradeitEmsEnvironments = TradeItEmsProductionEnv) {
+        if self.apiKey == nil {
```
jsom: Sure. It's a little more obvious but also adds a variable - I think it's ok either way.

TradeItIosTicketSDK2/TradeItSDK.swift
mitochondrion: Nope.  No errors.

PR #86, opened 2016-12-12
Fix up order price type/expiration error

jsom:


TradeItIosTicketSDK2/TradeItLauncher.swift
```diff
@@ -2,7 +2,7 @@
     let linkBrokerUIFlow = TradeItLinkBrokerUIFlow()
     let tradingUIFlow = TradeItTradingUIFlow()
     let viewControllerProvider = TradeItViewControllerProvider()
-    let deviceManager = TradeItDeviceManager()
+    var deviceManager = TradeItDeviceManager()
```
jsom: @mitochondrion had to make this `var` so the test could replace it with a fake. Don't feel good about that though. I could create some provider but it seems messy too.

TradeItIosTicketSDK2/TradeItSDK.swift
```diff
@@ -4,15 +4,15 @@
     private static var configured = false
     public static let launcher = TradeItLauncher()
 
-    private static var _linkedBrokerManager: TradeItLinkedBrokerManager?
+    internal static var _linkedBrokerManager: TradeItLinkedBrokerManager?
```
jsom: Had to make these internal so that the `@testable` tag will allow access. I think this is bad too. We should discuss workaround.

PR #93, opened 2016-12-27
Make session public

puchog:
Make session public in order to be able to access to the access token


puchog:
Hi guys, I make this modifications on the visibility of session object because on our project we need the user token to let the server handle a couple of actions (retrieving info and buying / selling stocks for now).
I hope this change of visibility no bothers anyone, but let me know your thoughts about this

PR #94, opened 2016-12-29
Renames TradeItAuthenticationDelegate -> TradeItOAuthDelegate

mitochondrion:
@jsom What you think about this?  I think the original name is very confusing, but will this change eff up current integrators?


jsom:
Nah I think you're correct to rename it. We just need to make sure to update the people using it.

PR #95, opened 2017-01-03
Account selection screen

jsom:


TradeItIosTicketSDK2/TradeItAccountSelectionViewController.swift
```diff
@@ -18,6 +20,7 @@ class TradeItAccountSelectionViewController: TradeItViewController, TradeItAccou
     override func viewWillAppear(_ animated: Bool) {
         super.viewWillAppear(animated)
 
+        self.promptLabel.text = promptText ?? "SELECT AN ACCOUNT FOR TRADING"
```
jsom: @mitochondrion not sure if I should make this a delegate method with a default? Seemed like overkill but I don't want any weirdness if the ViewController is reused. Not sure if that's worth covering.

PR #97, opened 2017-01-09
WIP: Cancelled security question should be in sessionError state

jsom:


TradeItIosTicketSDK2/TradeItLinkedBroker.swift
```diff
@@ -51,7 +51,7 @@ import PromiseKit
                             handler(TradeItErrorResult(
                                 title: "Authentication failed",
                                 message: "The security question was canceled.",
-                                code: .brokerAuthenticationError
+                                code: .sessionError
```
jsom: Setting it to `.sessionError` means that when `authenticateIfNeeded` is called again it will try to create a session. Before it was just showing an error.

PR #98, opened 2017-01-10
Fix unit tests after Yahoo merge

jsom:


TradeItIosTicketSDK2Tests/TradeItLinkedBrokerCacheSpec.swift
```diff
@@ -31,7 +31,7 @@ class TradeItLinkedBrokerCacheSpec: QuickSpec {
                 expect(self.defaults.dictionary(forKey: "LINKED_BROKER_CACHE")).to(beNil())
 
                 cache = TradeItLinkedBrokerCache()
```
mitochondrion: Do we still need this local var?

jsom: Yes, we use it later to setup data

PR #101, opened 2017-01-11
Use broker returned by oauth response

jsom:


TradeItIosEmsApi/TradeItAuthLinkResult.h
```diff
@@ -13,6 +13,7 @@
 
 @property (nullable) NSString *userId;
 @property (nullable) NSString *userToken;
+@property (nullable) NSString *broker;
```
jsom: We return `null` for non oauth flows -- should I leave this on here or move it to the `TradeItOAuthLinkResult` that inherits from this class? I don't think there's any problem leaving it here but it might be confusing later.

mitochondrion: Move to `TradeItOAuthLinkResult`.

TradeItIosTicketSDK2/TradeItLinkedBrokerManager.swift
```diff
@@ -73,15 +69,6 @@ import PromiseKit
     public func completeOAuth(withOAuthVerifier oAuthVerifier: String,
                               onSuccess: @escaping (_ linkedBroker: TradeItLinkedBroker) -> Void,
                               onFailure: @escaping (TradeItErrorResult) -> Void) -> Void {
-        guard self.currentOAuthBroker != nil else {
```
mitochondrion: Should we guard that result.broker is not null or empty?

PR #102, opened 2017-01-13
Fixes TradeItTradingUIFlow to only set trading account if not already set and only one account available

mitochondrion:
Noticed this bug and put in a fix. Please confirm this is correct behavior.


guillaumedebavelaere:
seems good to me

PR #103, opened 2017-01-16
Fixes Account Selection screen refresh to display relink alert only if authentication fails due to needing relink

mitochondrion:
@jsom @guillaumedebavelaere


mitochondrion:
Looked at this more closely with James. Seems that `showRelinkError` already checks internally, so this PR is not needed.

PR #106, opened 2017-02-07
Fix unowned LinkedBroker reference

jsom:
Potential crash if a user unlinks a broker, and the garbage collector cleans it up before the user goes back to the trading screen.

### DIFF
--- a/ExampleApp/ExampleViewController.swift
+++ b/ExampleApp/ExampleViewController.swift
@@ -139,6 +139,19 @@ class ExampleViewController: UIViewController, UITableViewDataSource, UITableVie
                 ]
             ),
             Section(
+                label: "Debugging",
+                actions: [
+                    Action(
+                        label: "deleteLinkedBrokers",
+                        action: deleteLinkedBrokers
+                    ),
+                    Action(
+                        label: "test",
+                        action: test
+                    )
+                ]
+            ),
+            Section(
                 label: "Themes",
                 actions: [
                     Action(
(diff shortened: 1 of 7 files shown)


ExampleApp/ExampleViewController.swift
```diff
@@ -126,6 +126,19 @@ class ExampleViewController: UIViewController, UITableViewDataSource, UITableVie
...
+                        label: "deleteLinkedBrokers",
+                        action: deleteLinkedBrokers
+                    ),
+                    Action(
+                        label: "test",
+                        action: test
+                    )
+                ]
+            ),
+            Section(
```
jsom: Moving this up - our UI tests fail if it's off-screen and swiping up/down is slow.

TradeItIosTicketSDK2/TradeItTradingTicketViewController.swift
```diff
@@ -207,12 +207,19 @@ class TradeItTradingTicketViewController: TradeItViewController, TradeItSymbolSe
     // MARK: Private - Order changed handlers
 
     private func accountSelected(linkedBrokerAccount: TradeItLinkedBrokerAccount) {
+        guard let linkedBroker = linkedBrokerAccount.linkedBroker else {
+            return self.alertManager.showError(TradeItErrorResult(
+                title: "Broker unlinked",
+                message: "The broker was unlinked. Please select another account.",
+                code: .systemError
```
jsom: Not a great code to give here. Thoughts?

PR #118, opened 2017-02-21
[#139297907] fix invalid api key issue

guillaumedebavelaere:
Linked to https://github.com/tradingticket/server/pull/948
As the session token is required, providing a default expired token, so if the user doesn't authenticate after launching from cache, he will get a session token expired.
Also, setting the session token only when the authenticate call is successful, otherwise it will reset it to null.

### DIFF
--- a/TradeItIosTicketSDK2/TradeItLinkedBrokerManager.swift
+++ b/TradeItIosTicketSDK2/TradeItLinkedBrokerManager.swift
@@ -325,6 +325,8 @@ import PromiseKit
 
     private func loadLinkedBrokerFromLinkedLogin(_ linkedLogin: TradeItLinkedLogin) -> TradeItLinkedBroker {
         let tradeItSession = sessionProvider.provide(connector: self.connector)
+        //provides a default token, so if the user doesn't authenticate before an other call, it will pass an expired token in order to get the session expired error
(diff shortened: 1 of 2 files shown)


TradeItIosTicketSDK2/TradeItLinkedBrokerManager.swift
```diff
@@ -325,6 +325,8 @@ import PromiseKit
 
     private func loadLinkedBrokerFromLinkedLogin(_ linkedLogin: TradeItLinkedLogin) -> TradeItLinkedBroker {
         let tradeItSession = sessionProvider.provide(connector: self.connector)
+        //provides a default token, so if the user doesn't authenticate before an other call, it will pass an expired token in order to get the session expired error
+        tradeItSession?.token = "dd61aa94fa094e6ab54fa4b31853bbd4"
```
jsom: In situations like this it's really helpful to use an actual dummy string to help with debugging rather than something that looks like it could be a valid token.

So maybe `trade-it-fetch-fresh-token`

guillaumedebavelaere:
A fix has been done server side: https://github.com/tradingticket/server/pull/948
But now it is returning that the session token is required, that is why we still need this fix in the SDK (As it wasn't passing always a session token before).

PR #120, opened 2017-02-21
adds empty lines at EOF on files which are missing one #trivial

dbgrandi:
I turned on the "missing newline at EOF" warning in Xcode recently and see a few here that I wanted to fix. :)


adamkaplan:
No more warnings! 🎉

adamkaplan:
History lesson: The reasoning for this dates back 28 years 
http://flash-gordon.me.uk/ansi.c.txt
Search for: `2.1.1.2 Translation phases` and item 2:
```
    A source file that is not empty shall end in a
    new-line character, which shall not be immediately preceded by a
    backslash character.
```

mitochondrion:
Wait what? That's a thing? Why would they do that??? I reflexively delete extraneous newlines when I see them. Besides being an offense to good taste, the extra carriage returns take disk space to store and replicate in the Github data center which increases the carbon footprint of our repos. Yet another way Apple is destroying the environment. smdh...

mitochondrion:
I offset the carbon emissions of the above comment having to be stored on the Github servers by not exhaling while typing it.

jsom:
I'm *sure* there was a good reason for this 28 years ago. Thank you for your contribution :us:

PR #123, opened 2017-02-22
[WIP][#139835777] fix UI tests

guillaumedebavelaere:


guillaumedebavelaere:
@jsom actually the UI tests and the unit tests are broken now with the new oAuth flow, I will work again on them

PR #153, opened 2017-03-29
Port trading ticket

jsom:


mitochondrion:
I think the square quantity input looks goofy with rounded corners. Maybe ask Ameer? Also the market price should have the timestamp of the quote (probably underneath as details text).

PR #156, opened 2017-03-30
Ak/handle oauth relink/141829615

mitochondrion:


TradeItIosTicketSDK2/TradeItAlertManager.swift
```diff
@@ -40,6 +40,19 @@ import UIKit
             )
         }
 
+        // TODO: ONLY ONE ALERT FOR BOTH????
```
mitochondrion: Ignore this for now...

PR #159, opened 2017-04-03
[#142867721] removes cache when unlinking

guillaumedebavelaere:


ExampleApp/ExampleViewController.swift
```diff
@@ -471,6 +471,9 @@ class ExampleViewController: UIViewController, UITableViewDataSource, UITableVie
 
         for linkedLogin in linkedLogins {
```
guillaumedebavelaere: Is there a reason why we are not using the linkedBrokerManager.unlinkBroker method here ?

jsom: I merged before I saw this comment. I don't think there is a reason but I think @mitochondrion wrote it. If you want to put in a PR to fix it and then maybe add the cache code within that too that would be good.

PR #162, opened 2017-04-07
OAuth uses SFSafariViewController instead of launching Safari

jsom:
Pros: Smoother UX
Cons: ~10mb higher memory footprint for the SDK even if the screen isn't used

### DIFF
--- a/TradeItIosTicketSDK2/TradeItLauncher.swift
+++ b/TradeItIosTicketSDK2/TradeItLauncher.swift
@@ -19,22 +19,29 @@ protocol OAuthCompletionListener {
     override internal init() {}
 
     public func handleOAuthCallback(
-        onViewController viewController: UIViewController,
+        onViewController safariViewController: UIViewController,
         oAuthCallbackUrl: URL
     ) {
         print("=====> handleOAuthCallback: \(oAuthCallbackUrl.absoluteString)")
 
         let oAuthCallbackUrlParser = TradeItOAuthCallbackUrlParser(oAuthCallbackUrl: oAuthCallbackUrl)
 
-        guard let originalViewController = safariViewController.presentingViewController?.presentingViewController else {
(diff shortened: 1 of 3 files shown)


TradeItIosTicketSDK2/TradeItLauncher.swift
```diff
@@ -19,19 +19,31 @@ protocol OAuthCompletionListener {
...
-            onSuccessfulLink: onSuccessfulLink
-        )
+        var parentViewController = safariViewController.presentingViewController
+        if parentViewController?.childViewControllers.first is TradeItSelectBrokerViewController {
+            parentViewController = parentViewController?.presentingViewController
+        }
+
+        guard let originalViewController = parentViewController else {
+            preconditionFailure("View hierarchy in unknown state.")
+        }
```
jsom: This seems a little magic. I'm not entirely sure if it can end up in some other state.

TradeItIosTicketSDK2/TradeItLauncher.swift
```diff
@@ -19,19 +19,31 @@ protocol OAuthCompletionListener {
...
         print("=====> handleOAuthCallback: \(oAuthCallbackUrl.absoluteString)")
 
         let oAuthCallbackUrlParser = TradeItOAuthCallbackUrlParser(oAuthCallbackUrl: oAuthCallbackUrl)
 
-        self.oAuthCompletionUIFlow.presentOAuthCompletionFlow(
-            fromViewController: viewController,
-            oAuthCallbackUrlParser: oAuthCallbackUrlParser,
-            onSuccessfulLink: onSuccessfulLink
-        )
+        var parentViewController = safariViewController.presentingViewController
```
mitochondrion: Add a comment about why you need to do this (relinking) as it is very confusing.

jsom:
The memory footprint issue is only on the simulator for some reason.

PR #163, opened 2017-04-12
Add NotificationCenter events for linking and unlinking that passes the Linked Broker

jsom:


TradeItIosTicketSDK2/TradeItOAuthCompletionViewController.swift
```diff
@@ -37,8 +37,11 @@ class TradeItOAuthCompletionViewController: TradeItViewController {
                         linkedBroker.refreshAccountBalances(onFinished: {
                             self.setSuccessState(forBroker: linkedBroker.brokerName)
                         })
+                        NotificationCenter.default.post(name: TradeItSDK.didLinkNotificationName, object: nil, userInfo: [
+                            "linkedBroker": linkedBroker
+                        ])
```
jsom: Important that this happens here because the Linked Broker is authenticated.

PR #165, opened 2017-04-19
Replace JSONModel in project with Cocoapod/Carthage dep

jsom:


jsom:
This is done but probably needs some heavy testing.

PR #171, opened 2017-05-23
Adds providedQuote to TradeItOrder to use it in the TradeItTradingTic…

guillaumedebavelaere:
…ketViewController

Adds a providedQuote property to the TradeItOrder so the partner can pass market data to the TradeItTradingTicketViewController (we do not handle market prices as yet from SGX)


jsom:
@mitochondrion can you review this. @guillaumedebavelaere I think Alex created a `quoteProvider` already

PR #174, opened 2017-05-26
TradeItLinkedBrokerAccount adds accountBaseCurrency

guillaumedebavelaere:


guillaumedebavelaere:
@jsom by default it will be "USD" if not in the keychain. It will be overridden the next time we fetch the account.

PR #175, opened 2017-05-30
[WIP] [#144666041] AlertManager.showRelinkError needs to handle 600 session…

guillaumedebavelaere:
… expiration


TradeItIosTicketSDK2/TradeItAlertManager.swift
```diff
@@ -61,6 +74,19 @@ import UIKit
                 showCancelAction: true,
                 onCancelActionTapped: onFinished
             )
+        case .sessionError?:
+            let title = error.shortMessage ?? ""
```
guillaumedebavelaere: The message returned by the server is the following:

title: `Could Not Complete Your Request`
message: `Your session has expired. Please try again`

Should we set in the SDK a more meaningful message for the user ?

PR #185, opened 2017-06-19
FX

jsom:


TradeItIosTicketSDK2/TradeItFxOrder.swift
```diff
@@ -0,0 +1,114 @@
...
+        fxOrderInfoInput.orderLegs = [orderLeg]
+
+        let request = TradeItFxPlaceOrderRequest()
+        request.accountNumber = order.linkedBrokerAccount?.accountNumber
+        request.fxOrderInfoInput = fxOrderInfoInput
+
+        return request
+    }
+
+//    private func action() -> String {
```
mitochondrion: Wut iz dis?

TradeItIosTicketSDK2/TradeItTradingTicketViewController.swift
```diff
@@ -48,6 +48,7 @@ class TradeItTradingTicketViewController: TradeItViewController, UITableViewData
         self.tableView.delegate = self
         self.tableView.dataSource = self
         self.tableView.tableFooterView = UIView()
+        TicketRow.registerNibCells(forTableView: self.tableView)
```
mitochondrion: Maybe move this call to some `UITableView` parent class or extension.

TradeItIosEmsApi/TradeItConnector.m
```diff
@@ -397,65 +397,64 @@ - (void)oAuthDeleteLink:(TradeItLinkedLogin *)linkedLogin {
...
+                  return;
+              }
+
+              NSMutableString *jsonResponse = [[NSMutableString alloc] initWithData:data encoding:NSUTF8StringEncoding];
+
+              TradeItResult *result = [TradeItRequestResultFactory buildResult:[ResultClass alloc] jsonString:jsonResponse];
+
+              if ([result.status isEqual:@"ERROR"]) {
+                  result = [TradeItRequestResultFactory buildResult:[TradeItErrorResult alloc] jsonString:jsonResponse];
+              }
```
jsom: There are 2 types of errors that can happen here:

1. Error from EMS: the `result` will be a `TradeItErrorResult` because the `ResultClass` could not be constructed. It will enter this `if` statement and build the `TradeItErrorResult` properly.
2. Error parsing JSON: the `result` will be a `TradeItErrorResult` for invalid JSON. It will get constructed again, but returns the same error.

jsom: Ah, this doesn't quite work. In the case of #2 it builds a `TradeItErrorResult` with `status = SUCCESS` which is wrong. Parse errors should not be interpreted again.

TradeItIosEmsApi/TradeItFxOrderInfoInput.h
```diff
@@ -0,0 +1,9 @@
+#include "TradeItFxOrderLeg.h"
+
+@interface TradeItFxOrderInfoInput : TradeItRequest
+
+@property (nonatomic, copy) NSString * _Nullable orderType;
+@property (nonatomic, copy) NSString * _Nullable orderExpiration;
+@property (nonatomic, copy) NSArray<TradeItFxOrderLeg> * _Nonnull orderLegs;
```
TheAntonioReyes: seems weird that this is nonnull while the other fields are nullable

PR #189, opened 2017-06-27
New welcome screen

jsom:


ExampleApp/AppDelegate.swift
```diff
@@ -4,15 +4,16 @@ import TradeItIosTicketSDK2
 
 @UIApplicationMain
 class AppDelegate: UIResponder, UIApplicationDelegate {
-    static let API_KEY = "tradeit-test-api-key" //"tradeit-fx-test-api-key"
+    static let API_KEY = "tradeit-fx-test-api-key" //"tradeit-test-api-key"
```
jsom: We don't want to do this

PR #191, opened 2017-06-27
support adding custom params when getting OAuth token

ihuxley:
Hey folk...

We need to add a custom param to the request for an oauth access token, I added this capability, please see if it is to your liking.

I know independently you're working on allowing us to create/sign your URLRequests, but that's not ideal for this case as the extra param is specific to the call launched in response to the deep link.

This may need tweaking for the new solution as it relies on mutating the URLRequest, but hopefully you get the jist of what we'd like, i.e. passing in the extra params specifically for getOAuthAccessToken.

--Iain.


ihuxley:
P.S. the re-serialization of the request is a little ugly, but I figure it wasn't worth mucking with TradeItRequestResultFactory at this point as I expect you're changing it.

jsom:
@ihuxley can I close this? It seems unneeded now.

ihuxley:
Oh, yes, this is unneeded since it turned out they can just go in the URL.

PR #194, opened 2017-07-07
Makes orderCapabilities optional

mitochondrion:


TradeItIosTicketSDK2/TradeItLinkedBroker.swift
```diff
@@ -207,7 +211,7 @@ import PromiseKit
                 balance: nil,
                 fxBalance: nil,
                 positions: [],
-                orderCapabilities: account.orderCapabilities as! [TradeItInstrumentOrderCapabilities],
+                orderCapabilities: account.orderCapabilities as? [TradeItInstrumentOrderCapabilities],
```
jsom: I think having the default value here makes sense so we don't have a bunch of optionals in the data model

PR #198, opened 2017-07-17
Add CDN logos

jsom:


TradeItIosTicketSDK2/TradeItSelectBrokerViewController.swift
```diff
@@ -219,4 +222,18 @@ class TradeItSelectBrokerViewController: CloseableViewController, UITableViewDel
 
         return cell
     }
+
+    private func getRemoteLogo(forBroker broker: TradeItBroker, imageView: UIImageView) -> Bool {
```
guillaumedebavelaere: Can we do some refactoring and put this code inside the `TradeItSDK.brokerLogoService.getLogo` method instead ?

jsom: Not really because `getLogo` is sync but this needs to be `async` and also needs the data from the `getBrokerList` call

jsom: Well... maybe. Let me think about it. We don't really want people to have to implement/overwrite this logic as far as I know.

PR #199, opened 2017-07-18
Ak/extract request provider

mitochondrion:


TradeItIosEmsApi/TradeItRequestResultFactory.m
```diff
@@ -69,6 +74,25 @@ + (NSString *)getHostForEnvironment:(TradeitEmsEnvironments)env {
...
+                                  @"Content-Type": @"application/json"
+                                  };
+
+        NSURLRequest *request = [TradeItRequestResultFactory.requestFactory buildRequestForUrl:url
+                                                                                    httpMethod:@"POST"
+                                                                                    parameters:requestParams
+                                                                                       headers:headers];
+
+        return (NSMutableURLRequest *)[request mutableCopy];
+    }
```
jsom: I still feel like we should have a default request factory. This isn't going to go away any time soon so it's useful for us to have single code paths that are more likely to be tested before we push out builds. If something broke unexpectedly here, we wouldn't see it until someone creating a `requestFactory` told us. Also is there anything we can/need to do to future proof this when if we rewrite this in Swift?

mitochondrion: I'll put in a story to refactor this stuff. Need to push out a build today.

PR #203, opened 2017-07-20
[WIP] Services to Swift

jsom:


TradeItIosEmsApi/TradeItConnector.m
```diff
@@ -94,12 +94,12 @@ - (void)getOAuthLoginPopupUrlForMobileWithBroker:(NSString *)broker
                                                                  broker:broker
                                                 interAppAddressCallback:[oAuthCallbackUrl absoluteString]];
 
-    NSMutableURLRequest *request = [TradeItRequestResultFactory buildJsonRequestForModel:oAuthLoginPopupUrlForMobileRequest
+    NSURLRequest *request = [TradeItRequestResultFactory buildJsonRequestForModel:oAuthLoginPopupUrlForMobileRequest
```
jsom: These should all be moved out to a Swift service for OAuth

TradeItIosTicketSDK2/TradeItBalanceService.swift
```diff
@@ -0,0 +1,33 @@
...
+    ) {
+        accountOverviewRequest.token = self.session.token
+
+        let request = TradeItRequestResultFactory.buildJsonRequest(
+            for: accountOverviewRequest,
+            emsAction: "balance/getAccountOverview",
+            environment: self.session.connector.environment
+        )
+
+        self.session.connector.send(request, targetClassType: TradeItAccountOverviewResult.self) { result in
```
jsom: This seems to be duplicated multiple times in the server. Might be a nice pattern to extract of:

1. Take data object
2. Construct request object
3. Pass to connector

TradeItIosTicketSDK2/TradeItConnector.swift
```diff
@@ -0,0 +1,53 @@
...
+
+                    let finalResult = result ?? TradeItErrorResult.error(withSystemMessage: "Server returned non TradeItResult")
+
+                    DispatchQueue.main.async {
+                        completionBlock(finalResult, json)
+                    }
+                } else {
+                    // TODO: Figure out what to do here? httpResponse failed or similar
+                    completionBlock(TradeItErrorResult.error(withSystemMessage: "Data Error"), nil)
+                }
```
jsom: I'm going to review this and try to make it have the exact same behaviour as previous version

TradeItIosEmsApi/TradeItConnector.m
```diff
@@ -94,12 +94,12 @@ - (void)getOAuthLoginPopupUrlForMobileWithBroker:(NSString *)broker
...
                                                                                emsAction:@"user/getOAuthLoginPopupUrlForMobile"
                                                                              environment:self.environment];
 
-    [self sendEMSRequest:request
-     withCompletionBlock:^(TradeItResult *tradeItResult, NSMutableString *jsonResponse) {
+    [self sendReturnJSON:request
```
guillaumedebavelaere: I don't see where this `sendReturnJSON` method is defined ?

jsom: https://github.com/tradingticket/TradeItIosTicketSDK2/pull/203/files#diff-8b668b73b631d78b7174d36515d27e69

jsom: I extracted this part to Swift. We should move the rest of these functions at some stage, but already I think this change is getting too big.

jsom: After this gets merged, if you're interested you could pull this out to a `TradeItOAuthService`

TradeItIosTicketSDK2/TradeIt.storyboard
```diff
@@ -1874,7 +1874,7 @@
                                 <nil key="highlightedColor"/>
                             </label>
                             <label opaque="NO" userInteractionEnabled="NO" contentMode="left" horizontalHuggingPriority="251" text="You can now trade from that account or view your portfolio to see performance, news, and edit settings." textAlignment="center" lineBreakMode="tailTruncation" numberOfLines="0" baselineAdjustment="alignBaselines" adjustsFontSizeToFit="NO" translatesAutoresizingMaskIntoConstraints="NO" id="QVk-PB-m4r" userLabel="Details Label">
-                                <rect key="frame" x="71" y="131" width="233" height="67"/>
+                                <rect key="frame" x="36" y="131" width="303" height="67"/>
```
guillaumedebavelaere: This change was intended ?

jsom: Yeah. Fixes the label on the link success screen being too thin.

TradeItIosTicketSDK2/TradeItMarketService.swift
```diff
@@ -6,93 +6,47 @@
...
-            withCompletionBlock: { result in
-                if let quotesResult = result as? TradeItQuotesResult,
-                    let quote = quotesResult.quotes?.first as? TradeItQuote {
-                    onSuccess(quote)
-                } else if let errorResult = result as? TradeItErrorResult {
-                    onFailure(errorResult)
-                } else {
-                    onFailure(TradeItErrorResult(title: "Market Data failed", message: "Fetching the quote failed. Please try again later."))
-                }
+        self.connector.send(request, targetClassType: TradeItQuotesResult.self, withCompletionBlock: { result in
```
jsom: TODO: Make sure this still works. I saw the market data missing on the trading screen.

guillaumedebavelaere: It was missing the apiKey in the request. Fixed

TradeItIosTicketSDK2/TradeItBalanceService.swift
guillaumedebavelaere: It seems it is already done ?

TradeItIosEmsApi/TradeItConnector.m
```diff
@@ -72,124 +70,30 @@ - (id)initWithApiKey:(NSString *)apiKey
...
-            NSLog(@"Could not fetch broker list; got error result: %@", result);
-            completionBlock(nil, nil);
-        }
-    }];
+    return self;
 }
 
-
+// TODO: Deprecated - should be using OAuth
 - (void)linkBrokerWithAuthenticationInfo:(TradeItAuthenticationInfo *)authInfo
```
guillaumedebavelaere: It isn't used anywhere except in the unit tests, can we delete this ?

jsom: Yes let's do it.

ExampleApp/ExampleViewController.swift
```diff
@@ -712,14 +712,22 @@ class ExampleViewController: UIViewController, UITableViewDataSource, UITableVie
         UserDefaults.standard.removePersistentDomain(forName: appDomain!)
 
         let connector = TradeItConnector(apiKey: AppDelegate.API_KEY, environment: AppDelegate.ENVIRONMENT, version: TradeItEmsApiVersion_2)
-
+        let oAuthService = TradeItOauthService(connector: connector)
```
jsom: I think rename to `TradeItOAuthService`

ExampleApp/ExampleViewController.swift
```diff
@@ -712,14 +712,22 @@ class ExampleViewController: UIViewController, UITableViewDataSource, UITableVie
...
-            connector.unlinkLogin(linkedLogin, localOnly: false) { result in
-                if let linkedBroker = TradeItSDK.linkedBrokerManager.linkedBrokers.filter({ $0.linkedLogin.userId == linkedLogin.userId }).first {
-                    TradeItSDK.linkedBrokerCache.remove(linkedBroker: linkedBroker)
+            oAuthService.unlinkLogin(
+                login: linkedLogin,
+                localOnly: false,
+                onSuccess: { result in
+                    if let linkedBroker = TradeItSDK.linkedBrokerManager.linkedBrokers.filter({ $0.linkedLogin.userId == linkedLogin.userId }).first {
+                            TradeItSDK.linkedBrokerCache.remove(linkedBroker: linkedBroker)
+                        }
```
jsom: Indentation is messed up here

TradeItIosEmsApi/TradeItConnector.m
guillaumedebavelaere: Actually it is used in a deprecated method:
```
@available(*, deprecated, message: "See documentation for supporting oAuth flow.")
    internal func linkBroker(
        authInfo: TradeItAuthenticationInfo,
        onSuccess: @escaping (_ linkedBroker: TradeItLinkedBroker) -> Void,
        onSecurityQuestion: @escaping (
            TradeItSecurityQuestionResult,
            _ submitAnswer: @escaping (String) -> Void,
            _ onCancelSecurityQuestion: @escaping () -> Void
        ) -> Void,
        onFailure: @escaping (TradeItErrorResult) -> Void
    ) {
        self.connector.linkBroker(with: authInfo) { authResult in
```

jsom: I think we can remove this method now as well.

jsom: It's marked as internal so no one should be able to use it. Let's get rid of it.

TradeItIosTicketSDK2/TradeItConnector.swift
```diff
@@ -0,0 +1,60 @@
...
+                    }
+
+                    let finalResult = result ?? TradeItErrorResult.error(withSystemMessage: "Server returned non TradeItResult")
+
+                    DispatchQueue.main.async {
+                        completionBlock(finalResult, json)
+                    }
+                } else {
+                    // TODO: Figure out what to do here? httpResponse failed or similar
+                    completionBlock(TradeItErrorResult.error(withSystemMessage: "Data Error"), nil)
```
jsom: @guillaumedebavelaere how do you think we should handle this?

jsom:
@guillaumedebavelaere can you take a look at my TODOs and tell me what you think we should do. Also are there still build warnings because of missing nullability specifiers? Let's try fix them as well

TradeItIosTicketSDK2/TradeItResultTransformer.swift
```diff
@@ -0,0 +1,23 @@
...
+            print("Server response: \(String(describing: json))")
+            return nil
+        } else if let result = result {
+            return result
+        } else {
+            print("TradeItResultTransformer Error")
+            print("Expected class: \(T.self)")
+            print("Server response: \(String(describing: json))")
+            return nil
+        }
```
jsom: We should maybe make the messages less aggressive here. It's not unusual for us to need to cast a few times so saying it's an Error is wrong. @guillaumedebavelaere what do you think would be useful to show?

jsom:
Also if you have time and want to: we should rename `TradeItRequestResultFactory` to `TradeItRequestFactory` and we could move that to Swift too

jsom:
@guillaumedebavelaere we can now get rid of `TradeItRequestResultFactory.buildResult`. You'll need to update `TradeItSymbolService`

TradeItIosTicketSDK2/TradeItConnector.swift
guillaumedebavelaere: We could set in the systemMessage the error message returned.
Example: I tested to connect to my local server which is not started, I got this error:
```
(lldb) po error?.localizedDescription
▿ Optional<String>
  - some : "Could not connect to the server."
```
But to be more obvious for developers, wouldn't it be useful to print out when there is a such error ?

TradeItIosTicketSDK2/TradeItResultTransformer.swift
guillaumedebavelaere: Not sure to understand this case: how it could return nil ? If we have an error, the error variable should be filled ?

TradeItIosTicketSDK2/TradeItConnector.swift
jsom: Print it out where? Technical error descriptions should never be shown to the user

guillaumedebavelaere: In the console I mean, for the developer.

guillaumedebavelaere:
@bluk16 when you have a chance, can you do a thorough test of this branch as we did pretty big changes to the underlying structure.

TradeItIosTicketSDK2/TradeItConnector.swift
```diff
@@ -34,27 +33,34 @@ public extension TradeItConnector {
...
         }
     }
+
+    private func processResponse(_ data: Data?, _ response: URLResponse?, _ error: Error?) -> (TradeItResult, String?) {
+        guard let httpResponse = response as? HTTPURLResponse else {
+            return (TradeItErrorResult.error(withSystemMessage: "Unable to cast response to HTTPUrlResponse."), nil)
+        }
+
+        guard httpResponse.statusCode == 200 else {
+            return (TradeItErrorResult.error(withSystemMessage: "Response status code: \(httpResponse.statusCode)."), nil)
```
guillaumedebavelaere: Maybe add the error message ? 
`error?.localizedDescription`

PR #205, opened 2017-07-21
[DNM] Color to reflect last price

niranjansenthilkumar:
reflecting if the price is up or down based on the day changed percentage.

There is a lag for the color reflection I can't seem to fix, and there is the issue of cost with the server constantly retrieving the day change value, so I am unsure on whether or not this should be a feature for the last price.


jsom:
Putting this on hold until we can get the day change.

niranjansenthilkumar:
revert bidask commit

PR #206, opened 2017-07-21
Fix Symbol Search Screen keyboard hiding content

bluk16:
1. Added Done button at the bottom of the keyboard. Select will hide keyboard
2. I have also hided the auto-suggestion from the keyboard, reason being that auto-suggestion should have no impact on user experience during symbol search. If you agree that is the case, the auto-suggestion is just using up space of the screen, which is significant for iphone 5 screen size.

screen shot 2017-07-21 at 11 33 25 am

### DIFF
--- a/TradeItIosTicketSDK2/TradeItSymbolSearchViewController.swift
+++ b/TradeItIosTicketSDK2/TradeItSymbolSearchViewController.swift
@@ -15,7 +15,7 @@ class TradeItSymbolSearchViewController: TradeItViewController, UITableViewDeleg
 
         self.activityIndicator.hidesWhenStopped = true
         setupSearchTextField()
-        
+
         TradeItSDK.adService.populate?(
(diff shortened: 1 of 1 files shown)


TradeItIosTicketSDK2/TradeItSymbolSearchViewController.swift
```diff
@@ -15,7 +15,6 @@ class TradeItSymbolSearchViewController: TradeItViewController, UITableViewDeleg
 
         self.activityIndicator.hidesWhenStopped = true
         setupSearchTextField()
-
```
mitochondrion: There should be vertical whitespace separating both explicit blocks of code and also implicit bocks of code to aid in visually organizing business logic.

bluk16: white space added

PR #209, opened 2017-07-26
[#149443315] Add onSuccess/onError callbacks to LinkedBrokerManager.u…

guillaumedebavelaere:
…nlinkBroker method


TradeItIosTicketSDK2/TradeItLinkedBrokerManager.swift
```diff
@@ -268,20 +268,34 @@ import PromiseKit
...
+                    NotificationCenter.default.post(
+                        name: TradeItSDK.didUnlinkNotificationName,
+                        object: nil,
+                        userInfo: [
+                            "linkedBroker": linkedBroker
+                        ]
+                    )
+                    onSuccess()
+                } else {
+                    let error = TradeItErrorResult(title: "UnlinkBroker error")
```
guillaumedebavelaere: Not sure what do we want in that case, should we just use onSuccess ?

ExampleApp/ExampleViewController.swift
```diff
@@ -713,14 +713,15 @@ class ExampleViewController: UIViewController, UITableViewDataSource, UITableVie
...
-            if let linkedBroker = TradeItSDK.linkedBrokerManager.linkedBrokers.filter({ $0.linkedLogin.userId == linkedLogin.userId }).first {
-                TradeItSDK.linkedBrokerCache.remove(linkedBroker: linkedBroker)
+            connector.unlinkLogin(linkedLogin, localOnly: false) { result in
+                if let linkedBroker = TradeItSDK.linkedBrokerManager.linkedBrokers.filter({ $0.linkedLogin.userId == linkedLogin.userId }).first {
+                    TradeItSDK.linkedBrokerCache.remove(linkedBroker: linkedBroker)
+                }
             }
         }
-
+
```
jsom: Why is xcode adding spaces here?

TradeItIosEmsApi/TradeItConnector.m
```diff
@@ -337,7 +339,8 @@ - (NSArray *)getLinkedLogins {
 }
 
 - (void)unlinkLogin:(TradeItLinkedLogin *)login
-          localOnly:(BOOL)localOnly {
+          localOnly:(BOOL)localOnly
```
jsom: Is `localOnly` ever used? I haven't seen it before. Was it added recently?

TradeItIosEmsApi/TradeItConnector.m
```diff
@@ -355,15 +358,21 @@ - (void)unlinkLogin:(TradeItLinkedLogin *)login
     [self.userDefaults setObject:accounts forKey:BROKER_LIST_KEYNAME];
 
     if (!localOnly) {
-        [self oAuthDeleteLink:login];
+        [self oAuthDeleteLink:login withCompletionBlock:completionBlock];
+    } else {
```
jsom: I think it reads clearer to reverse these so `if (localOnly) {` instead of having the `!` case first

TradeItIosTicketSDK2/TradeItLinkedBrokerManager.swift
jsom: Yeah not sure - @mitochondrion ?

jsom: I think the `title` is shown to the user so we should give it a nicer title. There might also be `TradeItErrorResult.systemError` helper.

TradeItIosTicketSDK2/TradeItAccountManagementViewController.swift
```diff
@@ -76,22 +76,28 @@ class TradeItAccountManagementViewController: TradeItViewController, TradeItAcco
...
+                                    fromViewController: navController,
+                                    showWelcomeScreen: true,
+                                    oAuthCallbackUrl: TradeItSDK.oAuthCallbackUrl
+                                )
+                            }
+                            navController.popViewController(animated: true)
+                        }
+                    },
+                    onFailure: { error in
+                        self.alertManager.showError(error, onViewController: self)
```
jsom: If for some reason we hit the case you mention below where `let error = TradeItErrorResult(title: "UnlinkBroker error")` then it's possible that we aren't handling the condition in the comments above either. I think in this case, if the broker doesn't exist locally anymore it should just be `onSuccess`

TradeItIosEmsApi/TradeItConnector.m
guillaumedebavelaere: It has been added recently by alex:
https://github.com/tradingticket/TradeItIosTicketSDK2/commit/d0a11f08255cd4d42371b4228cc4715cc1c4a9cb#diff-aeb438906a8062e23d46cc61ab15c791

mitochondrion: When the Yahoo apps sync from their backend, you don't want to call `/oAuthDelete` because the user could have signed in with a different Yahoo account and you don't want to unlink all the linked brokers associated with their other account.

TradeItIosTicketSDK2/TradeItAccountManagementViewController.swift
mitochondrion: Not sure what you mean by this?

mitochondrion: Fixed it so it will not delete local data if the remote call fails.

TradeItIosEmsApi/TradeItResult.h
```diff
@@ -39,5 +39,6 @@
  */
 @property (nullable, copy) NSArray<Optional> *longMessages;
 
+- (BOOL)isSuccessful;
```
jsom: lel

PR #213, opened 2017-07-31
[#149302069] Restore TOS, Privacy and Help to Select Your Broker Screen

guillaumedebavelaere:
Does this need to be done for Yahoo too ?


jsom:
@mitochondrion Yahoo?

@guillaumedebavelaere you should be able to delete the old Select a Broker screen from the Storyboard now. Also were the original link URLs not working? We should add redirect rules for the marketing site because there will still be older SDK versions in the wild.

guillaumedebavelaere:
@jsom only https://www.trade.it/faq is missing a 301 redirect. I asked Nico, Okapi is going to add it.

TradeItIosTicketSDK2/TradeItYahooOAuthCompletionUIFlow.swift
```diff
@@ -54,6 +54,7 @@ class TradeItYahooOAuthCompletionUIFlow: NSObject, TradeItYahooOAuthCompletionVi
                 onNavigationController: navController,
                 asRootViewController: true,
                 showWelcomeScreen: false,
+                hideOpenAccountButton: false,
```
jsom: Is this even used in the Yahoo screens?

TradeItIosTicketSDK2/TradeItLinkBrokerUIFlow.swift
```diff
@@ -46,10 +49,12 @@ class TradeItLinkBrokerUIFlow: NSObject, TradeItWelcomeViewControllerDelegate, L
         onNavigationController navController: UINavigationController,
         asRootViewController: Bool,
         showWelcomeScreen: Bool,
+        hideOpenAccountButton: Bool,
```
jsom: Seems confusing to have called this one `hide` when the one above is a `show`

ExampleApp/ExampleViewController.swift
```diff
@@ -213,6 +213,16 @@ class ExampleViewController: UIViewController, UITableViewDataSource, UITableVie
                         }
                     ),
                     Action(
+                        label: "Hide Open account button",
```
jsom: I don't think we need to add this in to the ExampleApp

TradeItIosTicketSDK2/TradeItYahooOAuthCompletionUIFlow.swift
guillaumedebavelaere: No (or not yet ?), but it is using the same linkBrokerUIFlow. I guess I could add a default value for the param.

TradeItIosTicketSDK2/TradeItLinkBrokerUIFlow.swift
guillaumedebavelaere: it was just to make the if more readable and I assumed by default it will be shown:
```
if hideOpenAccountButton {
     self.openNewAccountButton.removeFromSuperview()
 }
```

jsom: Yeah that's a good reason but I think in this case the person calling the function is more likely to get confused with the `show` `hide` interface than us in the code below

guillaumedebavelaere:
@jsom I changed hideOpenAccountButton -> showOpenAccountButton, add default value and remove the action from the ExampleApp.
Yahoo hasn't answered yet. Alex said we can finish this story and we will open a new one if needed.
Let me know if it is good to merge.

PR #215, opened 2017-08-07
[WIP] Remove themes

jsom:


ExampleApp/ExampleViewController.swift
```diff
@@ -100,35 +100,6 @@ class ExampleViewController: UIViewController, UITableViewDataSource, UITableVie
                 ]
             ),
             Section(
-                label: "Themes",
```
jsom: Probably need to add a demo for `UIAppearance`

jsom:
Closing because `UIAppearance` unfortunately is not as flexible.

PR #217, opened 2017-08-08
Injectable accounts

jsom:


TradeItIosTicketSDK2/TradeItLinkedBrokerManager.swift
```diff
@@ -524,20 +417,33 @@ import PromiseKit
...
 
         if let linkedLogin = linkedLogin {
             let linkedBroker = self.loadLinkedBrokerFromLinkedLogin(linkedLogin)
-            if userIdUserTokenBroker.isLinkActivationPending {
+            linkedBroker.accounts = linkedBrokerData.accounts.map { accountData in
+                return TradeItLinkedBrokerAccount(
+                    linkedBroker: linkedBroker,
+                    accountName: accountData.name,
+                    accountNumber: accountData.number,
+                    accountIndex: "",
```
jsom: Yahoo should just ignore this property as far as I know, so default this to empty string here. Not sure if needs comment?

TradeItIosTicketSDK2/TradeItLinkedBrokerManager.swift
```diff
@@ -622,25 +528,43 @@ import PromiseKit
     }
 }
 
-@objc public class UserIdUserTokenBroker: NSObject {
+@objc public class LinkedBrokerData: NSObject {
```
jsom: wat?

PR #222, opened 2017-08-14
Sync accounts

jsom:


TradeItIosTicketSDK2/TradeItLinkedBrokerManager.swift
```diff
@@ -313,15 +301,15 @@ import PromiseKit
         // Add missing linkedBrokers
         let localUserIds = self.linkedBrokers.flatMap { $0.linkedLogin.userId }
         let remoteLinkedBrokersToAdd = remoteLinkedBrokers.filter { !localUserIds.contains($0.userId) }
-        
+
         remoteLinkedBrokersToAdd.forEach { remoteBrokerData in
-            injectBroker(
+            self.saveLinkedBrokerToKeychain(
```
jsom: These were confusing. I thought initially these were async calls and was about to promisify everything but appears that it's actually sync.

TradeItIosTicketSDK2/TradeItLinkedBrokerManager.swift
```diff
@@ -331,10 +319,38 @@ import PromiseKit
...
+                syncAccounts(localLinkedBroker: localLinkedBroker, remoteLinkedBroker: remoteLinkedBroker)
+            }
+        }
+
         onFinished()
     }
 
+    private func syncAccounts(localLinkedBroker: TradeItLinkedBroker, remoteLinkedBroker: LinkedBrokerData) {
+        // Add missing accounts
+        let localAccountNumbers = localLinkedBroker.accounts.map { $0.accountNumber }
```
guillaumedebavelaere: isn't `flatMap` ?

TradeItIosTicketSDK2/TradeItLinkedBrokerManager.swift
```diff
@@ -331,10 +319,38 @@ import PromiseKit
...
             self.removeBroker(linkedBroker: linkedBrokerToRemove)
         }
-        
+
+        // Sync accounts
+        self.linkedBrokers.forEach { localLinkedBroker in
+            remoteLinkedBrokers.first { remoteLinkedBroker in
+                localLinkedBroker.linkedLogin.userId == remoteLinkedBroker.userId
+            }.flatMap { remoteLinkedBroker in
+                syncAccounts(localLinkedBroker: localLinkedBroker, remoteLinkedBroker: remoteLinkedBroker)
```
guillaumedebavelaere: We want also to update userToken (see comments in here : https://github.com/tradingticket/AndroidSDK/pull/63)

jsom: @guillaumedebavelaere are you sure? That means we'll need to read from the keychain every time. Yahoo plans to run this method every 30 seconds so it'll be very heavy.

jsom: @guillaumedebavelaere I'm going to leave this as is for now. Yahoo doesn't need to check the `userToken`

PR #223, opened 2017-08-15
Fixing error on the broker in the portfolio screen when it is a syste…

guillaumedebavelaere:
…m error


guillaumedebavelaere:
Closing, we are going to keep this for the fidelity demo but need a new story to handle error per account: https://www.pivotaltracker.com/story/show/150264712

PR #224, opened 2017-08-16
[#150299876] Pentest - create a new session with cache disabled

guillaumedebavelaere:


TradeItIosTicketSDK2/TradeItConnector.swift
```diff
@@ -31,7 +31,11 @@ internal extension TradeItConnector {
         withCompletionBlock completionBlock: @escaping (TradeItResult, String?) -> Void
     ) {
         DispatchQueue.global(qos: .userInitiated).async {
-            let session = URLSession.shared
+            let config = URLSessionConfiguration.default
+            config.requestCachePolicy = .reloadIgnoringLocalCacheData
+            config.urlCache = nil
+            let session = URLSession.init(configuration: config)
+
```
jsom: Can you do some research and see if this only needs to be created once? If so, we shouldn't create it every request. Also can you confirm that it won't appear in `Cache.db`

guillaumedebavelaere:
@jsom I checked, it doesn't appear in the Cache.db now.

guillaumedebavelaere:
@bluk16 can you do a thorough test of this branch ? Thanks

PR #227, opened 2017-08-17
Gd/encrypt cache with keychain/150332789

guillaumedebavelaere:


TradeItIosTicketSDK2/SerializedLinkedBroker.h
```diff
@@ -0,0 +1,21 @@
...
+//  Copyright © 2017 TradeIt. All rights reserved.
+//
+
+#import <JSONModel/JSONModel.h>
+#import "SerializedLinkedBroker.h"
+#import "SerializedLinkedBrokerAccount.h"
+
+@interface SerializedLinkedBroker : JSONModel
+
+@property (nonatomic) NSArray<SerializedLinkedBrokerAccount*> <SerializedLinkedBrokerAccount> *accounts;
```
jsom: This may not be the case: but could we use any of the existing JSONModels we have instead of creating new models to serialize?

TradeItIosTicketSDK2/TradeItIosTicketSDK2.h
```diff
@@ -4,5 +4,9 @@
 #import "TradeItIosEmsApiLib.h"
 #import "YFTopGradientView.h"
 #import "YFLinearGradientView.h"
+#import "SerializedLinkedBroker.h"
```
jsom: Maybe rename these to `CachedLinkedBroker` etc to make it clearer what these are used for.

TradeItIosTicketSDK2/TradeItLinkedBrokerCache.swift
```diff
@@ -1,178 +1,126 @@
...
-
-        if linkedBroker.isAccountLinkDelayedError {
-            linkedBroker.error = TradeItErrorResult(
-                title: "Activation In Progress",
-                message: "Your \(linkedBroker.brokerName) link is being activated which can take up to two business days. Check back soon.",
-                code: TradeItErrorCode.accountNotAvailable
-            )
+        
+        let serializedLinkedBroker = SerializedLinkedBroker(string: json, error: &error)
+        if let error = error {
```
jsom: Could you reuse the `TradeItResultTransformer` here?

guillaumedebavelaere:
@json is it good to merge ?

jsom:
@mitochondrion wrote the caching so we should get his opinion. We should get his opinion and also mention that it's going to lose the `enabled` flag to see what he thinks.

TradeItIosTicketSDK2/TradeItLinkedBrokerCache.swift
```diff
@@ -1,62 +1,51 @@
...
         guard let linkedBroker = linkedBroker else { return }
-
-        var linkedBrokerCache = userDefaults.dictionary(forKey: LINKED_BROKER_CACHE_KEY) as? SerializedLinkedBrokers ?? SerializedLinkedBrokers()
-
-        let serializedLinkedBroker = serialize(linkedBroker: linkedBroker)
-
-        linkedBrokerCache[linkedBroker.linkedLogin.userId] = serializedLinkedBroker
-
-        self.userDefaults.set(linkedBrokerCache, forKey: LINKED_BROKER_CACHE_KEY)
+        let key = LINKED_BROKER_CACHE_KEY+linkedBroker.linkedLogin.userId
```
mitochondrion: `linkedBroker.linkedLogin.userId` can be changed to `linkedBroker.userId`. There is a convenience computed property getter on `TradeItLinkedBroker`.

TradeItIosTicketSDK2/TradeItLinkedBrokerCache.swift
```diff
@@ -1,62 +1,51 @@
...
-            let accounts = deserialize(
-                serializedAccounts: serializedAccounts,
-                forLinkedBroker: linkedBroker
+        let key = LINKED_BROKER_CACHE_KEY+linkedBroker.linkedLogin.userId
+        guard let json = TradeItKeychain.getStringForKey(key)
+             ,let cachedLinkedBroker = TradeItResultTransformer.transform(targetClassType: CachedLinkedBroker.self, json: json)
+            else { return}
+        
+        linkedBroker.accounts = cachedLinkedBroker.accounts.map { cachedAccount in
+            let account = TradeItLinkedBrokerAccount(
```
mitochondrion: Can we use JSONModel to deserialize the JSON directly into the nested account/balance objects instead of creating them mapping all the fields over manually?

TradeItIosTicketSDK2/TradeItLinkedBrokerCache.swift
```diff
@@ -67,112 +56,58 @@ class TradeItLinkedBrokerCache {
...
+        self.balanceLastUpdated = linkedBrokerAccount.balanceLastUpdated
+    }
+}
 
-                serializedAccount[FX_BALANCE_BUYING_POWER_KEY] = buyingPowerBaseCurrency
-            }
+extension CachedAccountOverview {
+    convenience init(balance: TradeItAccountOverview?) {
+        self.init()
+        self.buyingPower = balance?.buyingPower
```
mitochondrion: Since we are serializing to JSON and don't need to create new keys in NSUserDefaults anymore to store new data, should we just cache all the balance fields instead of just `buyingPower`?

TradeItIosTicketSDK2/TradeItLinkedBrokerCache.swift
guillaumedebavelaere: If we are ok to store all the properties ?

PR #236, opened 2017-08-28
Ak/user event notification

mitochondrion:


ExampleApp/AppDelegate.swift
```diff
@@ -67,6 +69,64 @@ class AppDelegate: UIResponder, UIApplicationDelegate {
...
+            name: TradeItNotification.Name.errorShown,
+            object: nil
+        )
+    }
+
+    func onErrorShownNotification(notification: Notification) {
+        let view = notification.userInfo?[TradeItNotification.UserInfoKey.view] ?? "NO KEY FOR VIEW"
+        let errorTitle = notification.userInfo?[TradeItNotification.UserInfoKey.errorTitle] ?? "NO KEY FOR ERROR CODE"
+        let errorMessage = notification.userInfo?[TradeItNotification.UserInfoKey.errorMessage] ?? "NO KEY FOR ERROR MESSAGE"
+        print("=====> ERROR SHOWN: VIEW: \(view), TITLE: \(errorTitle), MESSAGE: \(errorMessage)")
```
jsom: Do we want to just pass the error object back here?

ExampleApp/Base.lproj/Main.storyboard
```diff
@@ -72,5 +72,53 @@
             </objects>
             <point key="canvasLocation" x="-678" y="-290"/>
         </scene>
+        <!--Example View Controller-->
+        <scene sceneID="uFO-j0-SXY">
```
jsom: Why is this VC new?

TradeItIosTicketSDK2/Constants.swift
```diff
@@ -0,0 +1,27 @@
...
+    public static let nameViewDidAppear = TradeItNotification.Name.viewDidAppear
+    public static let nameErrorShown = TradeItNotification.Name.errorShown
+
+    // MARK: Notification UserInfo keys
+    public static let userInfoKeyView = TradeItNotification.UserInfoKey.view
+    public static let userInfoKeyErrorTitle = TradeItNotification.UserInfoKey.errorTitle
+    public static let userInfoKeyErrorMessage = TradeItNotification.UserInfoKey.errorMessage
+}
+
+public struct TradeItNotification {
```
jsom: I don't think Swift structs are available in obj-c

TradeItIosTicketSDK2/TradeItYahooViewController.swift
```diff
@@ -19,4 +19,16 @@ class TradeItYahooViewController: CloseableViewController {
             containerView.bottomAnchor.constraint(equalTo: self.view.bottomAnchor)
         ])
     }
+
+    override func viewDidAppear(_ animated: Bool) {
+        super.viewDidAppear(animated)
+
+        NotificationCenter.default.post(
```
jsom: Nice

ExampleApp/AppDelegate.swift
mitochondrion: That makes sense.

ExampleApp/Base.lproj/Main.storyboard
mitochondrion: To fix some silliness that occurs when reusing the storyboard nib for advanced options screen.

TradeItIosTicketSDK2/Constants.swift
mitochondrion: See the @objc class above.

PR #237, opened 2017-08-28
Gd/set title selection screens/150596693

guillaumedebavelaere:


jsom:
@guillaumedebavelaere let's hold on merging this. I think the title might look squished because they are sorta long.

PR #242, opened 2017-08-29
[#150618342] Report CIMB - display broker long name

guillaumedebavelaere:


TradeItIosEmsApi/TradeItConnector.m
```diff
@@ -173,6 +181,7 @@ - (NSArray *)getLinkedLogins {
     for (NSDictionary *account in linkedAccounts) {
         [accountsToReturn addObject:[[TradeItLinkedLogin alloc] initWithLabel:account[@"label"]
                                                                        broker:account[@"broker"]
+                                                                       brokerLongName:account[@"brokerLongName"] ?: account[@"broker"]
```
jsom: This defaults the value, but will that have any quirky UI side effects? Will this record be saved at a later time with the correct long name?

TradeItIosTicketSDK2/TradeItRequestFactory.swift
```diff
@@ -31,7 +31,7 @@ class TradeItRequestFactory: NSObject {
     private static var envToHostDict: [TradeitEmsEnvironments: String] = [
         TradeItEmsProductionEnv : "https://ems.tradingticket.com/",
         TradeItEmsTestEnv: "https://ems.qa.tradingticket.com/",
-        TradeItEmsLocalEnv: "https://localhost:8443/"
+        TradeItEmsLocalEnv: "http://localhost:8080/"
```
jsom: Is this change permanent/intentional?

jsom:
Are there benefits to this approach of storing the long name vs using the value we get back from the broker list? What if the broker name changes (as per OptionsHouse -> E*Trade)?

TradeItIosEmsApi/TradeItConnector.m
guillaumedebavelaere: The broker longname is passed when linking, so it will be updated the next time the user relink for existing accounts.

TradeItIosTicketSDK2/TradeItRequestFactory.swift
guillaumedebavelaere: It was for testing locally, because I was unable to test it with https. If you know how to test with https locally, can you detail the procedure ?

guillaumedebavelaere:
If the broker name changes it will be updated on the next relink (notice it is currently the case for the short name too). 
It seems more complicated if we wanted to use the broker long name from the list, to get it on the completion screen (or do you have an idea in mind ?)

jsom:
@guillaumedebavelaere can't the long name be retrieved from the `getBrokerList` call which is cached on SDK `configure` which will already be loaded by the time someone gets to the OAuthCompletion screen

jsom:
I don't know if it's necessarily easier, but it's 1 less key to manage/keep fresh.

guillaumedebavelaere:
I find it more complex:
We would have to do something like:
```
getAvailableBrokers(onSuccess: { brokers in
                    let brokerLongName = brokers.filter {$0.shortName == linkedBroker.brokerName}
                }, onFailure: { error in
                
                })
```
vs just get the brokerLongName from the response.
Actually the broker shortName/longname wasn't updated after a relink, I just changed it

jsom:
Cool. Alright yeah that's fine! Merging.

PR #246, opened 2017-08-31
[#150685848] Report CIMB - implement orderCapabilities on equity ticket

guillaumedebavelaere:


TradeItIosTicketSDK2/TradeItPreviewTradeOrderDetails+Helpers.swift
```diff
@@ -1,31 +1,122 @@
...
+        default:
+            return "unknown"
+        }
+    }
+    
+    public init?(rawValue: RawValue) {
+        self.init(value: rawValue)
+    }
+    
+    public init(value: RawValue?) {
```
jsom: Do these work for Obj-C users?

jsom:
How will this work for anyone that has integrated their own custom UI and are trying to use the enums to set values?

TradeItIosTicketSDK2/TradeItPreviewTradeOrderDetails+Helpers.swift
guillaumedebavelaere: No, in Objective C, you can use only the enum `TradeItOrderActionBuy`, ...

guillaumedebavelaere:
What do you mean ? 
Example to get the label:
```
self.orderCapabilities?.labelFor(field: .priceTypes, value: self.order.type.rawValue)
```

jsom:
Some partners construct their own order objects with:

```swift
let order = TradeItOrder()
order.type = .limit
```

It seems like with this change we will no longer support this. I don't think it's a blocker, but we should do some combination of:

1. Put this on a new minor version
2. Contact partners that may do this to tell them that they need to upgrade and instruct them on how to use the order caps object

guillaumedebavelaere:
```
let order = TradeItOrder()
order.type = .limit
```

This is still working. But yes, if they implement their own ticket, we should write a documentation on how do use the order capabilities.

jsom:
Oh ok - so if they pass a limit order to the order object and submit it to a broker that doesn't support it, it will just fail server side?

guillaumedebavelaere:
Yes

jsom:
@guillaumedebavelaere can you rebase and merge

guillaumedebavelaere:
@jsom @mitochondrion rebased. Could you review the last commit, I had to fix conflicts on some code you changed.

TradeItIosTicketSDK2/Constants.swift
```diff
@@ -67,6 +67,7 @@ public struct TradeItNotification {
         case trading
         case preview
         case submitted
+        case selectActionType
```
jsom: Can you add this to the objective-c mapping above in `TradeItNotificationConstants`

guillaumedebavelaere: @jsom done

PR #247, opened 2017-09-01
[147723953] AllOrderStatus

guillaumedebavelaere:


TradeItIosTicketSDK2/TradeItBalanceService.swift
```diff
@@ -1,4 +1,4 @@
-@objc public class TradeItBalanceService: NSObject {
+internal class TradeItBalanceService: NSObject {
```
jsom: Why this change?

jsom:
Some minor comments but otherwise looks good.

PR #250, opened 2017-09-06
Add ActivityIndicatorFactory

jsom:


TradeItIosTicketSDK2/ActivityIndicatorFactory.swift
```diff
@@ -0,0 +1,14 @@
...
+@objc public protocol ActivityIndicatorFactory {
+    func build(frame: CGRect) -> UIView
+}
+
+class DefaultActivityIndicatorFactory: ActivityIndicatorFactory {
+    func build(frame: CGRect) -> UIView {
+        let view = UIActivityIndicatorView(frame: frame)
+        view.color = UIColor.gray
+        view.startAnimating()
+        return view
```
jsom: `UIActivityIndicatorView` doesn't adjust the spinner size based on what frame you pass in so it just shows a small spinner anyway.

TradeItIosTicketSDK2/TradeItSDK.swift
```diff
@@ -1,9 +1,21 @@
...
+    // MARK: Settable properties
+
+    public static var theme: TradeItTheme = TradeItTheme.light()
+    public static var isPortfolioEnabled = true
+    public static var isAdServiceEnabled = false
+    public static var userCountryCode: String? // CountryCode matching standard: ISO3166 alpha-2. Used for managing broker availability.
+    public static var adService: AdService = DefaultAdService()
+    public static var welcomeScreenHeadlineText: String = "Link your broker account to enable:"
+    public static var featuredBrokerLabelText: String = "SPONSORED BROKER"
+    public static var activityViewFactory: ActivityIndicatorFactory = DefaultActivityIndicatorFactory()
```
jsom: Should I add this to the config calls?

guillaumedebavelaere: If we add this to the trade it screens, yes, otherwise it could be confusing ?

guillaumedebavelaere:
There is also an authenticateCall in the ticket screen where we potentially show a spinner and when we preview and place the order, do we want to use it there ?

jsom:
Still waiting to hear back from Yahoo about `MBProgressHud` spinner

ihuxley:
LGTM, thanks!

jsom:
Merging and we can revisit other spinners later if needed.

PR #251, opened 2017-09-07
[#149410359][#149421881][#149422315][#149422047][#149271291] Orders screen

guillaumedebavelaere:


TradeItIosTicketSDK2/TradeItOrdersTableViewManager.swift
```diff
@@ -0,0 +1,240 @@
...
+        self.refreshControl = refreshControl
+    }
+    
+    /**
+     * This is to split orders in order to have a specific section for group orders
+    **/
+    private func getSplittedOrdersArray(orders: [TradeItOrderStatusDetails]) -> [[TradeItOrderStatusDetails]]{
+        return orders.reduce([[]], { splittedArrays, order in
+            var splittedArraysTmp = splittedArrays
+            let lastResult: [TradeItOrderStatusDetails] = splittedArrays[(splittedArrays.endIndex - 1)]
```
jsom: Why not use `.last`?

TradeItIosTicketSDK2/TradeItOrdersTableViewManager.swift
```diff
@@ -0,0 +1,240 @@
...
+        )
+        TradeItThemeConfigurator.configure(view: refreshControl)
+        tableView.addSubview(refreshControl)
+        self.refreshControl = refreshControl
+    }
+    
+    /**
+     * This is to split orders in order to have a specific section for group orders
+    **/
+    private func getSplittedOrdersArray(orders: [TradeItOrderStatusDetails]) -> [[TradeItOrderStatusDetails]]{
```
jsom: Is there a better name we can think of to use here? I'm not 100% sure what it's doing by something like `groupOrderLegs` or `collateOrders`?

TradeItIosTicketSDK2/TradeItOrdersTableViewManager.swift
```diff
@@ -0,0 +1,240 @@
...
+        return orders.reduce([[]], { splittedArrays, order in
+            var splittedArraysTmp = splittedArrays
+            let lastResult: [TradeItOrderStatusDetails] = splittedArrays[(splittedArrays.endIndex - 1)]
+            
+            if !order.isGroupOrder() && !lastResult.contains(order) { // this is not a group order, we can append the order
+                splittedArraysTmp[(splittedArraysTmp.endIndex - 1)].append(order)
```
jsom: `.last`

TradeItIosTicketSDK2/TradeItOrdersTableViewManager.swift
```diff
@@ -0,0 +1,240 @@
...
+            self.orderSectionPresenters.append(
+                OrderSectionPresenter(
+                    orders: orders,
+                    title: title,
+                    isGroupOrder: isGroupOrder
+                )
+            )
+        }
+    }
+
```
jsom: This function is a little bit hard to follow - `forEach` usually indicates that there's is a functional way to do this. Maybe think about moving some of those conditions and `title` decisions to a separate function. Using `map` is probably going to clean this up as well.

TradeItIosTicketSDK2/TradeItOrdersViewController.swift
```diff
@@ -0,0 +1,94 @@
...
+        }
+        
+        func ordersPromise() -> Promise<Void> {
+            return Promise<Void> { fulfill, reject in
+                linkedBrokerAccount.getAllOrderStatus(
+                    onSuccess: { orders in
+                        self.ordersTableViewManager?.updateOrders(orders)
+                        fulfill()
+                    },
+                    onFailure: { error in
```
jsom: `onFailure: reject`

TradeItIosTicketSDK2/TradeItOrdersViewController.swift
```diff
@@ -0,0 +1,94 @@
...
+                    onFailure: reject
+                )
+            }
+        }
+        
+        func ordersPromise() -> Promise<Void> {
+            return Promise<Void> { fulfill, reject in
+                linkedBrokerAccount.getAllOrderStatus(
+                    onSuccess: { orders in
+                        self.ordersTableViewManager?.updateOrders(orders)
```
jsom: Because this is a side effect rather than the request for data, it might be best to move this below in to its own `then` so set this to `onSuccess: fulfill` and change the promise type to accept the orders. Then below add:
```swift
authenticatePromise().then { _ in
    return ordersPromise()
}.then { orders in
    self.ordersTableViewManager?.updateOrders(orders)
}.catch //... etc
```

jsom:
Looks good @guillaumedebavelaere ! Just a few comments

jsom:
@guillaumedebavelaere are there plans to add a Trade button to the Orders screen?

guillaumedebavelaere:
@jsom No directly,  but there is a trade button when there are no results. Otherwise there is a story to change the portfolio button on the confirmation screen by a button to go to the orders screen.

TradeItIosTicketSDK2/TradeItOrdersTableViewManager.swift
guillaumedebavelaere: Actually last return an immutable variable, so ?I can't use it here, but I will for above

TradeItIosTicketSDK2/TradeItOrdersViewController.swift
jsom: I realise I didn't do this in the recent changes I made to portfolio. I'll try fix it there.

jsom:
Merging this as I'm building a Swift 4 branch

PR #257, opened 2017-09-13
If last price is missing use quote

jsom:


TradeItIosTicketSDK2/TradeItPortfolioAccountDetailsViewController.swift
```diff
@@ -76,32 +76,66 @@ class TradeItPortfolioAccountDetailsViewController: TradeItViewController, Trade
...
-            return when(fulfilled: accountOverviewPromise, positionsPromise)
-        }.catch { error in
-            print(error)
-        }.always {
-            onRefreshComplete()
+    private func quotesPromise(portfolioPositions: [TradeItPortfolioPosition]) -> Promise<[TradeItPortfolioPosition]> {
+        let symbols = portfolioPositions
+            .filter { $0.position?.lastPrice == nil }
+            .map { $0.position?.symbol }
+            .flatMap { $0 }
```
guillaumedebavelaere: Is it possible when position is nil to end with a nil object in the `symbols` array ?

guillaumedebavelaere:
I wrote just a question otherwise it looks good

TradeItIosTicketSDK2/TradeItPortfolioAccountDetailsViewController.swift
jsom: The `flatMap` should remove it but I'll double check

jsom: I can probably combine the `map` and `flatMap` actually

PR #260, opened 2017-09-15
[DNM] Swift 4

jsom:


TradeItIosTicketSDK2/TradeItLinkedBrokerManager.swift
```diff
@@ -274,7 +274,7 @@ import PromiseKit
         onFinished: @escaping () -> Void
     ) {
         // Add missing linkedBrokers
-        let localUserIds = self.linkedBrokers.flatMap { $0.linkedLogin.userId }
+        let localUserIds: [String] = self.linkedBrokers.flatMap { $0.linkedLogin.userId }
```
jsom: Without defining the type, Swift 4 would interpret this as `flatMap` all the way down including the strings.

TradeItIosTicketSDK2/TradeItStepperInputTableViewCell.swift
```diff
@@ -31,7 +31,7 @@ class TradeItStepperInputTableViewCell: UITableViewCell {
         }
     }
 
-    func dismissKeyboard() {
+    @objc func dismissKeyboard() {
```
guillaumedebavelaere: Why do we need `@objc` here ?

TradeItIosTicketSDK2/TradeItPortfolioAccountDetailsViewController.swift
```diff
@@ -67,7 +67,7 @@ class TradeItPortfolioAccountDetailsViewController: TradeItViewController, Trade
                 cacheResult: true,
                 onSuccess: { _ in
                     self.tableViewManager.updateAccount(withAccount: linkedBrokerAccount)
-                    fulfill()
+                    fulfill(())
```
guillaumedebavelaere: What do the extra parenthesis mean ?

TradeItIosTicketSDK2/TradeItStepperInputTableViewCell.swift
jsom: Functions passed to `#selector` need to be available to Obj-C. In Swift 3 it was inferred, but Swift 4 they are removing that inference.

PR #262, opened 2017-09-18
[#150492095] Update trade confirm screen - change portfolio button by…

guillaumedebavelaere:
… orders status

### DIFF
--- a/TradeItIosTicketSDK2/TradeItSDK.swift
+++ b/TradeItIosTicketSDK2/TradeItSDK.swift
@@ -4,7 +4,7 @@ import UIKit
     // MARK: Settable properties
 
     public static var theme: TradeItTheme = TradeItTheme.light()
-    public static var isOrderStatusEnabled = true
+    public static var isPortfolioEnabled = true
     public static var isAdServiceEnabled = false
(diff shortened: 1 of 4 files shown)


TradeItIosTicketSDK2/TradeItSDK.swift
```diff
@@ -4,7 +4,7 @@ import UIKit
     // MARK: Settable properties
 
     public static var theme: TradeItTheme = TradeItTheme.light()
-    public static var isPortfolioEnabled = true
+    public static var isOrderStatusEnabled = true
```
jsom: Why is this removed? It seemed from the story that we want both the Portfolio and the Order Status button available but configurable.

guillaumedebavelaere: @jsom we replaced the portfolio view button by the order view button, and the `isPortfolioEnabled ` is not used anywhere else.

jsom: It is used by partners - they will have to modify their code to use this new boolean

jsom: I think we should leave the flag as is but use it to also cover this case. If someone disables the portfolio, that should also mean disabling the orders so essentially it all means the same.

PR #264, opened 2017-09-18
Fix Y! issue for status bar increase on iOS X

jsom:


TradeItIosTicketSDK2/TradeItYahooViewController.swift
```diff
@@ -3,6 +3,7 @@ import UIKit
 class TradeItYahooViewController: CloseableViewController {
     override func viewDidLoad() {
         super.viewDidLoad()
+        guard let yahooNavigationController = self.navigationController as? TradeItYahooNavigationController else { return }
```
jsom: Not 100% sure what to do here - default to a number? It should never happen but it might.

PR #266, opened 2017-09-19
[#151190515] Order screen - fix dummy 700

guillaumedebavelaere:


TradeItIosTicketSDK2/TradeItOrdersViewController.swift
```diff
@@ -86,8 +88,6 @@ class TradeItOrdersViewController: TradeItViewController, TradeItOrdersTableDele
                 withLinkedBroker: self.linkedBrokerAccount?.linkedBroker,
                 onViewController: self
             )
```
guillaumedebavelaere: There is an issue with dummy 600 though. We will get an alert with session expired - cancel - refresh. With the refresh, the screen won't be updated.
If we add the refresh in an onFinished, this will loop in case of error:
```
self.alertManager.showAlertWithAction(
                error: error,
                withLinkedBroker: self.linkedBrokerAccount?.linkedBroker,
                onViewController: self,
                onFinished: {
                    self.refreshRequested(onRefreshComplete: onRefreshComplete)
                }
            )
```
Thoughts ?

guillaumedebavelaere: Maybe we should call showAlertWithAction only is it is a 700 error ?

bluk16:
The alert is changed. When I select `refresh`, it is not redirected to the OAuth login screen.

screen shot 2017-09-19 at 9 32 30 am

guillaumedebavelaere:
This is the 600 error. As I mentioned, the refresh doesn't refresh the screen. You will need to pull to refresh to trigger the 700 error.

bluk16:
Thanks. I see what @gdebavelaere said here now.

TradeItIosTicketSDK2/TradeItOrdersViewController.swift
jsom: Is `onFinished` called if a user clicks Cancel?

guillaumedebavelaere: @jsom Yes it is

jsom:
@guillaumedebavelaere I think we leave this as is for now. I believe it's behaving the same way that the rest of the SDK does. I don't want to change `showAlertWithAction` until we have decent test coverage on it because I remember we did it in the past and it led to other edge cases breaking.

guillaumedebavelaere:
Ok, good to merge then ? @bluk16 the dummy 700 error is ok ?

PR #269, opened 2017-09-19
Add Trading with singapore symbol to the example app for testing CIMB

guillaumedebavelaere:
I added this to the example app in order to test more easily CIMB (because our symbol selection screen doesn't search for Singapore symbols).

If you think it is not mergeable, could you install this version on Nico's phone ? He needs to pass a trade and then test the security question when cancelling an order from the orders screen for CIMB.

Thanks


jsom:
Just did this manually so that he could test - had to set the API key as well anyway.

PR #275, opened 2017-10-09
[#151724711] Fix crash order screen  - TradeItOrderStatusDetailsPrese…

guillaumedebavelaere:
…nter - add safe check on the fills table


jsom:
👍 Looks good - anywhere else on the orders page that might have similar unsafe array accesses?

PR #276, opened 2017-10-09
Gd/ios doc/149157729

guillaumedebavelaere:


docs/_includes/_Sidebar.html
```diff
@@ -0,0 +1,29 @@
+<div id="toc" class="tocify">
```
guillaumedebavelaere: I took the menu from the wiki:
```
Home
Installation
Usage
Launching TradeIt screens
Custom integration
Testing
User Action Tracking
FAQ
```

But there are actually pages that are not in this menu:
```
Ad Configuration
eToro
Event API
Linking a broker with OAuth
Special Case Integration
```

@jsom @mitochondrion Do these need to be added to the menu ? Otherwise the user won't be able to find them if there are no internal links

jsom: @guillaumedebavelaere can you add the `Ad Configuration` one - otherwise the rest are pretty custom cases so don't need to add them for everyone to see. Make sure they're still available through the Github wiki though.

PR #277, opened 2017-10-09
Hide rows for missing balance fields

jsom:


TradeItIosTicketSDK2/TradeItPortfolioAccountDetailsTableViewManager.swift
```diff
@@ -7,7 +7,16 @@ class TradeItPortfolioAccountDetailsTableViewManager: NSObject, UITableViewDeleg
         case count
     }
 
+    private enum ACCOUNT_DETAIL_ROWS: Int {
+        case totalValue = 0
```
guillaumedebavelaere: Why do we assign a value only for `totalValue `? Should we assign all of them or none ?

TradeItIosTicketSDK2/TradeItPortfolioAccountDetailsTableViewManager.swift
```diff
@@ -135,7 +153,41 @@ class TradeItPortfolioAccountDetailsTableViewManager: NSObject, UITableViewDeleg
...
+                return self.accountDetailCell(
+                    forTableView: tableView,
+                    title: "Day return",
+                    value: presenter.getFormattedDayReturnWithPercentage(),
+                    valueColor: presenter.getDayReturnChangeColor()
+                )
+            case .buyingPower:
+                return self.accountDetailCell(
+                    forTableView: tableView,
+                    title: "Buying power",
```
guillaumedebavelaere: Is it possible to use the buyingPowerLabel returned by the server ? (It could be different for CIMB)
There should be an existing method on the presenter.

TradeItIosTicketSDK2/TradeItPortfolioAccountDetailsTableViewManager.swift
```diff
@@ -135,7 +153,41 @@ class TradeItPortfolioAccountDetailsTableViewManager: NSObject, UITableViewDeleg
...
+                )
+            case .availableCash:
+                return self.accountDetailCell(
+                    forTableView: tableView,
+                    title: "Available cash",
+                    value: presenter.getFormattedAvailableCash()
+                )
+            }
         } else {
             let position = self.positions?[indexPath.row]
```
guillaumedebavelaere: We could add a safe array access check

TradeItIosTicketSDK2/TradeItPortfolioAccountDetailsTableViewManager.swift
```diff
@@ -176,22 +228,28 @@ class TradeItPortfolioAccountDetailsTableViewManager: NSObject, UITableViewDeleg
     }
 
     func tableView(_ tableView: UITableView, heightForRowAt indexPath: IndexPath) -> CGFloat {
+        if indexPath.section == SECTIONS.accountDetails.rawValue && indexPath.row > ACCOUNT_DETAIL_ROWS.totalValue.rawValue {
+            return 32
```
guillaumedebavelaere: We could create a constant variable for this number

TradeItIosTicketSDK2/TradeItPortfolioAccountDetailsTableViewManager.swift
jsom: `enum: Int` auto-increments -- sometimes `1` is used as a starting value so I think it's idiomatic to define the first value.

PR #279, opened 2017-10-16
Gd/portfolio remove day return empty/151932242

guillaumedebavelaere:
```
Please remove Day/Total Return rows in the portfolio stock detail page (when user clicks into a stock in the portfolio) when there is not data. Currently, Fidelity does not provide this for us and it might be confusing for the user if there is just a blank spot next to day/total return.
```

@jsom @mitochondrion I used UIStackView. It works but I get some error message in the storyboard I don't understand. Could we pair on this ?

### DIFF
--- a/TradeItIosTicketSDK2/TradeIt.storyboard
+++ b/TradeItIosTicketSDK2/TradeIt.storyboard
@@ -5,7 +5,7 @@
     </device>
     <dependencies>
         <deployment identifier="iOS"/>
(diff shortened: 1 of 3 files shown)


guillaumedebavelaere:
@jsom could we pair on this today ?

guillaumedebavelaere:
@jsom I removed all the constraint errors and used the nested stackview we talked about. However I had to put a height size on the nested stackview to make it working ...

jsom:
Cool! Well done. I think this is a much better way of doing it than the original way I had done it 😄

PR #280, opened 2017-10-17
[#152040583] Transactions

guillaumedebavelaere:


TradeItIosEmsApi/TradeItIosEmsApiLib.h
```diff
@@ -88,6 +88,11 @@
 #import "TradeItPriceInfo.h"
 #import "TradeItCancelOrderRequest.h"
 
+// Transactions
+#import "TradeItTransactionsHistoryRequest.h"
+#import "TradeItTransactionsHistoryResult.h"
+#import "TradeItTransactionHistory.h"
```
jsom: I think this should just be called `TradeItTransaction`

TradeItIosEmsApi/TradeItTransactionHistory.h
```diff
@@ -0,0 +1,25 @@
...
+
+@property (nonatomic, copy, nullable) NSNumber<Optional>  * amount;
+
+@property (nonatomic, copy, nullable) NSString<Optional>  * action;
+
+@property (nonatomic, copy, nullable) NSString<Optional>  * type;
+
+@property (nonatomic, copy, nullable) NSString<Optional>  * id;
+
+@property (nonatomic, copy, nullable) NSNumber<Optional>  * quantity;
```
jsom: Usually it's written `NSNumber<Optional> *quantity`

TradeItIosEmsApi/TradeItTransactionsHistoryResult.h
```diff
@@ -0,0 +1,13 @@
...
+
+@class TradeItTransactionHistory;
+
+@protocol TradeItTransactionHistory
+
+@end
+
+@interface TradeItTransactionsHistoryResult : TradeItResult
+
+@property(nonatomic, copy, nullable) NSArray<TradeItTransactionHistory*> <Optional, TradeItTransactionHistory>  * transactionHistoryDetailsList;
```
jsom: Is the array actually nullable? Why?

TradeItIosTicketSDK2/TradeItTransactionService.swift
```diff
@@ -0,0 +1,37 @@
...
+    
+    init(session: TradeItSession) {
+        self.session = session
+    }
+    
+    func getTransactionsHistory(
+        _ data: TradeItTransactionsHistoryRequest,
+        onSuccess: @escaping (TradeItTransactionsHistoryResult) -> Void,
+        onFailure: @escaping (TradeItErrorResult) -> Void
+        ) {
```
jsom: Indentation is off

TradeItIosEmsApi/TradeItTransactionsHistoryRequest.h
```diff
@@ -0,0 +1,8 @@
+#import "TradeItRequest.h"
+
+@interface TradeItTransactionsHistoryRequest : TradeItRequest
+
+    @property (nonatomic, copy, nullable) NSString * token;
+    @property (nonatomic, copy, nullable) NSString * accountNumber;
```
jsom: Remove indentation

TradeItIosEmsApi/TradeItIosEmsApiLib.h
jsom: Have these been added to both the SDK and SDK-Carthage targets?

guillaumedebavelaere: > I think this should just be called TradeItTransaction

Ok, I will rename it

> Have these been added to both the SDK and SDK-Carthage targets?

Yes

TradeItIosEmsApi/TradeItTransactionsHistoryResult.h
guillaumedebavelaere: No, but I put all optional to avoid blowing up in case the server returns null

jsom: The server should not return `null` here ever. Do we have an E2E test that asserts this returns an array?

jsom: @mitochondrion thoughts?

guillaumedebavelaere:
@jsom @mitochondrion is it good to merge ?

PR #281, opened 2017-10-17
Gd/pentest jailbreak/151293378

guillaumedebavelaere:


TradeItIosTicketSDK2/TradeItSDK.swift
```diff
@@ -133,6 +133,8 @@ import UIKit
             return
         }
 
+        precondition(!TradeItDeviceManager.isDeviceJailBroken(), "ERROR: the device is jailbroken, couldn't load the TradeItSDK !")
```
jsom: This will just totally crash the app. Do we have to crash? Can we just deny access to any TradeIt functionality? Partners don't want us to ever be the cause of their app crashing.

TheAntonioReyes: @guillaumedebavelaere can we just put this in front of the linking so that anything trying to open the oAuth url will just issue an error message and never actually request a URL from our server?

TradeItIosTicketSDK2/TradeItConnector.swift
```diff
@@ -32,6 +32,16 @@ internal extension TradeItConnector {
         _ request: URLRequest,
         withCompletionBlock completionBlock: @escaping (TradeItResult, String?) -> Void
     ) {
+        guard !TradeItSDK.isDeviceJailbroken else {
+            completionBlock(
+                TradeItErrorResult(
```
guillaumedebavelaere: @TheAntonioReyes , @jsom This will return a TradeItErroResult for any request if the device is jailbroken.

guillaumedebavelaere:
This works on our test device (jailbroken), we get the error. But it would be nice if we could test if it still works on a non jailbroken iPhone device.

TheAntonioReyes:
@bluk16 can you help ^^^

TradeItIosTicketSDK2/TradeItConnector.swift
jsom: @TheAntonioReyes will this satisfy the pen test issue?

TradeItIosTicketSDK2/TradeItDeviceManager.swift
```diff
@@ -79,4 +79,48 @@ class TradeItDeviceManager {
...
+            || FileManager.default.fileExists(atPath: "/private/var/mobile/Library/SBSettings/Themes")
+            || FileManager.default.fileExists(atPath: "/private/var/stash")
+            || FileManager.default.fileExists(atPath: "/private/var/tmp/cydia.log")
+            || FileManager.default.fileExists(atPath: "/System/Library/LaunchDaemons/com.ikey.bbot.plist")
+            || FileManager.default.fileExists(atPath: "/System/Library/LaunchDaemons/com.saurik.Cydia.Startup.plist")
+            || FileManager.default.fileExists(atPath: "/bin/bash")
+            || FileManager.default.fileExists(atPath: "/usr/bin/sshd")
+            || FileManager.default.fileExists(atPath: "/etc/apt")
+            || FileManager.default.fileExists(atPath: "/usr/libexec/sftp-server")
+            || FileManager.default.fileExists(atPath: "/usr/sbin/sshd")
```
jsom: Feels weird to call `fileExists` so verbosely every time. Can you add a private static array of these values:
```
if JAIL_BREAK_FILES.any(FileManager.default.fileExists) || UIApplication.shared.canOpenURL(URL(string:"cydia://package/com.example.package")!) {
  return true
}
```

TradeItIosTicketSDK2/TradeItSDK.swift
```diff
@@ -133,6 +140,8 @@ import UIKit
             return
         }
 
+        self._isDeviceJailbroken = TradeItDeviceManager.isDeviceJailBroken()
```
jsom: Is it not possible to move this to the `get`? It does not rely on any of the parameters passed in

TradeItIosTicketSDK2/TradeItSession.swift
```diff
@@ -55,10 +55,10 @@ public class TradeItSession {
         if let authenticationResult = TradeItResultTransformer.transform(targetClassType: TradeItAuthenticationResult.self, json: json) {
             self.token = authenticationResult.token
             return authenticationResult
-        } else if let securityQuestionResult = TradeItResultTransformer.transform(targetClassType: TradeItSecurityQuestionResult.self, json: json) {
+        } else if let securityQuestionResult = authenticationResult as? TradeItSecurityQuestionResult {
```
jsom: Where does this get transformed now?

TradeItIosTicketSDK2/TradeItConnector.swift
TheAntonioReyes: take out the space and bang " !" and we should be good

TradeItIosTicketSDK2/TradeItDeviceManager.swift
guillaumedebavelaere: There doesn't seem to be an `any` method.
Is this ok to do that  : `!JAIL_BREAK_FILES.filter(FileManager.default.fileExists).isEmpty`

TradeItIosTicketSDK2/TradeItSession.swift
guillaumedebavelaere: https://github.com/tradingticket/TradeItIosTicketSDK2/blob/2e50f46fb87183de63d5fc44e7475f9ef5363405/TradeItIosTicketSDK2/TradeItConnector.swift#L78

bluk16:
Tested on non-jailbreak iphone 6, worked as expected.

PR #285, opened 2017-10-26
[#151532538] SDK TradingTicket - Show N/A for stocks with no quotes

guillaumedebavelaere:


TradeItIosTicketSDK2/TradeItQuotePresenter.swift
```diff
@@ -12,7 +12,7 @@ class TradeItQuotePresenter: NSObject {
     }
 
     func formatCurrency(_ value: NSNumber?) -> String? {
-        guard let value = value else { return TradeItPresenter.MISSING_DATA_PLACEHOLDER }
+        guard let value = value, value != 0 else { return TradeItPresenter.MISSING_DATA_PLACEHOLDER }
```
jsom: @TheAntonioReyes any idea why last price would be returned as `0` for an Xignite quote? Is it safe to just ignore these values or are they meaningful?

TheAntonioReyes: they are meaningful, we found $0.00 pricing for some of the FranklinTemplton funds because they were very illiquid. I think it's fine to show the placeholder though in these scenarios

PR #287, opened 2017-10-27
Gd/orders screen group order status/152129903

guillaumedebavelaere:


guillaumedebavelaere:
@mitochondrion tested, ok

TradeItIosTicketSDK2/TradeItOrdersTableViewManager.swift
```diff
@@ -157,6 +177,78 @@ class TradeItOrdersTableViewManager: NSObject, UITableViewDelegate, UITableViewD
         self.refreshControl = refreshControl
     }
     
+    private func addSwipeToCancel(header: UITableViewCell, section: Int) {
```
mitochondrion: This doesn't have the same behavior as when swiping non-header rows (looks different, doesn't bounce, can't swipe all the way to activate, etc.). It also breaks on older versions of iOS (like the test phone or if you load an older iOS version in the simulator). Can we implement the grouped order parents as rows instead of headers?

PR #299, opened 2017-11-10
Gd/yahoo orders screen/151530890

guillaumedebavelaere:
Created on top of #287

### DIFF
--- a/TradeItIosTicketSDK2/TradeItViewController.swift
+++ b/TradeItIosTicketSDK2/TradeItViewController.swift
@@ -1,9 +1,32 @@
 import UIKit
 
 class TradeItViewController: CloseableViewController {
+
+    public var enableThemeOnLoad: Bool = true
+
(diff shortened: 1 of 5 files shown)


TradeItIosTicketSDK2/TradeItViewController.swift
```diff
@@ -1,9 +1,32 @@
...
     override func viewDidLoad() {
         super.viewDidLoad()
-
-        TradeItThemeConfigurator.configure(view: self.view)
+        if enableThemeOnLoad {
+            TradeItThemeConfigurator.configure(view: self.view)
+        }
+    }
+    
+    func enableCustomNavController() {
```
mitochondrion: I thought we were going to change `TradeItYahooViewController` to inherit from this and set `enableThemeOnLoad = false` so that we could remove the duplicate code in `enableCustomNavController`?

PR #302, opened 2017-11-11
Add checks for account security type support to all trading tickets

mitochondrion:


TradeItIosTicketSDK2/TradeItFxTradingTicketViewController.swift
```diff
@@ -66,6 +66,46 @@ class TradeItFxTradingTicketViewController: TradeItViewController, UITableViewDa
 
     override func viewWillAppear(_ animated: Bool) {
         super.viewWillAppear(animated)
+
+        let isChosenAccountEnabled = self.order.linkedBrokerAccount?.isEnabled ?? false
```
guillaumedebavelaere: What about something like:
```swift
guard let isChosenAccountEnabled = self.order.linkedBrokerAccount?.isEnabled, isChosenAccountEnabled else {
     self.showAccountSelection()
     return
}
```
?

guillaumedebavelaere: Also what is the use case for this ? I thought we were already only displaying enabled accounts on the account selection screen ?

TradeItIosTicketSDK2/TradeItFxTradingTicketViewController.swift
```diff
@@ -66,6 +66,46 @@ class TradeItFxTradingTicketViewController: TradeItViewController, UITableViewDa
...
+        if (!isChosenAccountEnabled) {
+            self.showAccountSelection()
+            return
+        }
+
+        // Check if selected account supports trading FX
+        TradeItSDK.linkedBrokerManager.getAvailableBrokers(
```
guillaumedebavelaere: Is it possible to use the orderCapabilities on the linkedBrokerAccount ?
```swift
guard let  _ = self.order.linkedBrokerAccount?.orderCapabilities(forInstrument: TradeItTradeInstrumentType.fx)` else {
//alert
}
```

TradeItIosTicketSDK2/TradeItTradingTicketViewController.swift
```diff
@@ -76,11 +76,43 @@ class TradeItTradingTicketViewController: TradeItViewController, UITableViewData
...
         let isChosenAccountEnabled = self.order.linkedBrokerAccount?.isEnabled ?? false
         if (!isChosenAccountEnabled) {
-            self.accountSelectionViewController.selectedLinkedBrokerAccount = self.order.linkedBrokerAccount
-            self.navigationController?.pushViewController(self.accountSelectionViewController, animated: true)
+            self.showAccountSelection()
             return
         }
 
+        // Check if selected account supports trading equities
+        TradeItSDK.linkedBrokerManager.getAvailableBrokers(
```
guillaumedebavelaere: Is it possible to use the orderCapabilities on the linkedBrokerAccount ?
```swift
guard let  _ = self.order.linkedBrokerAccount?.orderCapabilities(forInstrument: TradeItTradeInstrumentType.equities)` else {
//alert
}
```

TradeItIosTicketSDK2/TradeItTradingTicketViewController.swift
```diff
@@ -76,11 +76,43 @@ class TradeItTradingTicketViewController: TradeItViewController, UITableViewData
 
         let isChosenAccountEnabled = self.order.linkedBrokerAccount?.isEnabled ?? false
```
guillaumedebavelaere: What about something like:
```swift
guard let isChosenAccountEnabled = self.order.linkedBrokerAccount?.isEnabled, isChosenAccountEnabled else {
     self.showAccountSelection()
     return
}
```
?

TradeItIosTicketSDK2/TradeItFxTradingTicketViewController.swift
mitochondrion: If you go to the account selection screen from the ticket and disable the chosen account and then navigate back to the ticket, the disabled account will still be selected. This will detect that scenario.

mitochondrion: Changing to:
```
guard self.order.linkedBrokerAccount?.isEnabled ?? false {
    self.showAccountSelection()
    return
}
```

TradeItIosTicketSDK2/TradeItFxTradingTicketViewController.swift
mitochondrion: I don't want to switch to that yet because it's not yet supported by the Linked Broker sync functionality.

TradeItIosTicketSDK2/TradeItFxTradingTicketViewController.swift
```diff
@@ -66,6 +66,45 @@ class TradeItFxTradingTicketViewController: TradeItViewController, UITableViewDa
...
+            onSuccess: { brokers in
+                guard let broker = brokers.first(
+                    where: { broker in
+                        return broker.shortName == self.order.linkedBrokerAccount?.linkedBroker?.brokerName
+                    }
+                ), broker.equityServices()?.supportsTrading == true else {
```
guillaumedebavelaere: Should be `broker.fxService` ?

TradeItIosTicketSDK2/TradeItFxTradingTicketViewController.swift
mitochondrion: Also I just looked at the server and we haven't even implemented this for equities!

guillaumedebavelaere: It should be implemented in the account response for both. 
The second endpoint `getFxOrderCapabilities` is when it depends on the ticker. Indeed there is no implementation for equities for that case.

guillaumedebavelaere: For equity, we should be able to do the check here: https://github.com/tradingticket/TradeItIosTicketSDK2/blob/ac2f996e98e2f0bec69614cfc9575fc494dc4838/TradeItIosTicketSDK2/TradeItTradingTicketViewController.swift#L313

As we authenticateIfNeeded when we load the ticket, we should get the order capabilities, even if it is not in the cache ?

TradeItIosTicketSDK2/TradeItTradingTicketViewController.swift
mitochondrion: We don't fetch orderCapabilities for equity trading and I don't think we even have implemented it yet in the API.

PR #307, opened 2017-11-21
[#152932975] Refactor Trading Ticket account checks to use order capabilities

guillaumedebavelaere:


jsom:
Now that we call the order capabilities every time the screen appears won't that slow the user down unnecessarily? If they're switching the order type, etc we don't need to call update order caps again. cc @mitochondrion as I believe you made this change on FX?

guillaumedebavelaere:
Is there another way than passing a flag we could call orderCapabilities only if the account has changed?

TradeItIosTicketSDK2/TradeItTradingTicketViewController.swift
```diff
@@ -81,46 +82,21 @@ class TradeItTradingTicketViewController: TradeItViewController, UITableViewData
...
-                        withTitle: "Unsupported Account",
-                        withMessage: "The selected account does not support trading stocks. Please choose another account.",
-                        withActionTitle: "OK",
-                        onAlertActionTapped: {
-                            self.delegate?.invalidAccountSelected(
-                                onTradingTicketViewController: self,
-                                withOrder: self.order
-                            )
-                        }
+        guard let _ = self.orderCapabilities else {
```
mitochondrion: This doesn't seem to be checking if equity trading is supported, just if orderCapabilities exists, which won't exist if we didn't call authenticate yet.

TradeItIosTicketSDK2/TradeItYahooTradingTicketViewController.swift
```diff
@@ -66,51 +65,7 @@ class TradeItYahooTradingTicketViewController: TradeItYahooViewController, UITab
...
-            }
-        )
-
-        self.reloadTicket()
-    }
-
-    override func viewDidAppear(_ animated: Bool) {
-        super.viewDidAppear(animated)
-        self.fireViewEventNotification(view: .trading, title: self.title)
+        self.updateOrderCapabilities()
```
mitochondrion: I thought we were going to separate this stuff so we don't fetch balances every time we come back to the screen...

guillaumedebavelaere: I haven't updated yet the Yahoo ticket. I wanted to confirm that was correct for the Trade it Ticket to update it.

TradeItIosTicketSDK2/TradeItTradingTicketViewController.swift
guillaumedebavelaere: This is set before. On `viewDidLoad` we call `updateOrderCapabilities` and in it: `self.orderCapabilities = (self.order.linkedBrokerAccount?.orderCapabilities.filter { $0.instrument == "equities" })?.first`
We set it also when we switch account.

guillaumedebavelaere:
@jsom Yahoo Trading Ticket updated

jsom:
Needs a rebase

guillaumedebavelaere:
@jsom rebased

PR #310, opened 2017-11-22
[#148168391] Transactions screen

guillaumedebavelaere:


TradeItIosTicketSDK2/TradeItTransactionTableViewCell.swift
```diff
@@ -0,0 +1,22 @@
...
+class TradeItTransactionTableViewCell: UITableViewCell {
+
+    @IBOutlet weak var transactionTypeLabel: UILabel!
+    @IBOutlet weak var timestampLabel: UILabel!
+    @IBOutlet weak var descriptionLabel: UILabel!
+    @IBOutlet weak var amountLabel: UILabel!
+    
+    override func awakeFromNib() {
+        super.awakeFromNib()
+    }
```
jsom: Need to add theming to these screens and cells

TradeItIosTicketSDK2/TradeItTransactionsViewController.swift
```diff
@@ -0,0 +1,97 @@
...
+                            securityQuestion,
+                            onViewController: self,
+                            onAnswerSecurityQuestion: answerSecurityQuestion,
+                            onCancelSecurityQuestion: cancelSecurityQuestion
+                        )
+                },
+                    onFailure: reject
+                )
+            }
+        }
```
jsom: I think we have this in a bunch of places now. Maybe we can extract it to an internal method on the `linkedBroker`?

jsom:
Looks good - needs to add theming though for all new UI

jsom:
There's also a merge conflict -- need to fix that on all of the Transactions branches

jsom:
Feel free to add separate story for the theming

guillaumedebavelaere:
Added a story for theming: https://www.pivotaltracker.com/story/show/153216556

TradeItIosTicketSDK2/TradeItTransactionsViewController.swift
guillaumedebavelaere: Because of the security question we have a reference to the alertManager and the view:
```
alertManager.promptUserToAnswerSecurityQuestion(
                        securityQuestion,
                        onViewController: self,
                        onAnswerSecurityQuestion: answerSecurityQuestion,
                        onCancelSecurityQuestion: cancelSecurityQuestion
                    )
```
How we could refactor this in the linkedBroker ?

guillaumedebavelaere:
conflict fixed

PR #311, opened 2017-11-22
[#148169577] Transaction detail screen

guillaumedebavelaere:
created on top of https://github.com/tradingticket/TradeItIosTicketSDK2/pull/310

### DIFF
--- a/TradeItIosTicketSDK2/TradeItTransactionDetailsViewController.swift
+++ b/TradeItIosTicketSDK2/TradeItTransactionDetailsViewController.swift
@@ -20,35 +20,22 @@ class TradeItTransactionDetailsViewController: TradeItViewController, UITableVie
         }
         self.transactionDetailTableView.delegate = self
         self.transactionDetailTableView.dataSource = self
-        self.presenter = TradeItTransactionPresenter(transaction, currencyCode: accountBaseCurrency)
-        self.title = self.presenter?.getTransactionTypeLabel()
-        if self.presenter?.getTransactionDescriptionLabel() != "" {
-            self.transactionDetailsRows.append(.description)
-        }
-        if self.presenter?.getTransactionActionLabel() != "" {
-            self.transactionDetailsRows.append(.action)
-        }
-        if self.presenter?.getTransactionQuantityLabel() != "" {
-            self.transactionDetailsRows.append(.quantity)
-        }
-        if self.presenter?.getTransactionSymbolLabel() != "" {
-            self.transactionDetailsRows.append(.symbol)
-        }
-        if self.presenter?.getTransactionPriceLabel() != "" {
-            self.transactionDetailsRows.append(.price)
-        }
-        if self.presenter?.getTransactionCommissionLabel() != "" {
-            self.transactionDetailsRows.append(.commission)
-        }
-        if self.presenter?.getAmountLabel() != "" {
-            self.transactionDetailsRows.append(.amount)
-        }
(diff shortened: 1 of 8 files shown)


TradeItIosTicketSDK2/TradeItTransactionDetailsViewController.swift
```diff
@@ -0,0 +1,70 @@
...
+            preconditionFailure("TradeItIosTicketSDK ERROR: TradeItTransactionDetailsViewController loaded without setting transaction.")
+        }
+        guard let accountBaseCurrency = self.accountBaseCurrency else {
+            preconditionFailure("TradeItIosTicketSDK ERROR: TradeItTransactionDetailsViewController loaded without setting accountBaseCurrency.")
+        }
+        self.transactionDetailTableView.delegate = self
+        self.transactionDetailTableView.dataSource = self
+        self.presenter = TradeItTransactionPresenter(transaction, currencyCode: accountBaseCurrency)
+        self.title = self.presenter?.getTransactionTypeLabel()
+        if self.presenter?.getTransactionDescriptionLabel() != "" {
```
jsom: @guillaumedebavelaere what if this returns null? It will still add the row. Is that intended?

TradeItIosTicketSDK2/TradeItTransactionPresenter.swift
```diff
@@ -41,7 +41,14 @@ class TradeItTransactionPresenter: NSObject {
         
         return "\(formatLabel(action)) \(quantity) \(symbol) @ \(price)"
     }
-    
+
+    func getTransactionDescriptionLabel() -> String {
+        guard self.transaction.typeEnum != .trade else {
+            return getDescriptionLabel()
+        }
+        return self.transaction.transactionDescription ?? ""
```
jsom: We should probably return a `String?` here so you don't use the defaults all the time. The `null` case has meaning in that you won't be adding the row to be displayed. Otherwise you need to do all those `!= ""` checks above which is strange.

TradeItIosTicketSDK2/TradeItTransactionDetailsViewController.swift
```diff
@@ -0,0 +1,70 @@
...
+        if self.presenter?.getAmountLabel() != "" {
+            self.transactionDetailsRows.append(.amount)
+        }
+        if self.presenter?.getTransactionDateLabel() != "" {
+            self.transactionDetailsRows.append(.date)
+        }
+        if self.presenter?.getTransactionIdLabel() != "" {
+            self.transactionDetailsRows.append(.id)
+        }
+        self.transactionDetailTableView.reloadData()
```
jsom: Might be nicer to create an array of tuples and filter it to get an array of rows to assign to `transactionDetailsRow`.

TradeItIosTicketSDK2/TradeItTransactionPresenter.swift
guillaumedebavelaere: We don't want to display if it is empty so I will have to check this in the if like:
`let transactionLabel = if self.presenter?.getTransactionActionLabel() && transactionLabel != ""` ?
Or do you have another idea ?

TradeItIosTicketSDK2/TradeItTransactionDetailsViewController.swift
guillaumedebavelaere: 👍

guillaumedebavelaere:
@jsom updated

PR #313, opened 2017-11-24
marginType -> userCanDisableMargin

jsom:


TradeItIosTicketSDK2/CachedLinkedBrokerAccount.m
```diff
@@ -2,4 +2,9 @@
 
 @implementation CachedLinkedBrokerAccount
 
++ (BOOL)propertyIsOptional:(NSString *)propertyName {
+    NSArray *optionalProperties = @[@"userCanDisableMargin", @"balanceLastUpdated", @"balance", @"fxBalance"];
+    return [optionalProperties containsObject:propertyName];
+}
```
jsom: Had to add this because any `CachedLinkedBrokerAccount` without `userCanDisableMargin` already saved would fail to deserialize. `BOOL` is a primitive and so you can't mark it as `Optional`. This is the workaround.

mitochondrion: Where is this used?

PR #317, opened 2017-12-07
[#153487833] Add fidelityPilot flag

guillaumedebavelaere:
Need to test when this one will be merged in QA: https://github.com/tradingticket/server/pull/1810


jsom:
@gdebavelaere what WIP items are needed here?

guillaumedebavelaere:
Just need to test when server code will be deployed in QA.
Also, I was thinking about your comment in the other PR.  For example, if there are some SDK fixes the partner wants but without having to upgrade to FidelityV1, we may need to create a separate version/branch?

PR #320, opened 2017-12-11
[#152438619] Add position description

guillaumedebavelaere:


TradeItIosEmsApi/TradeItPosition.h
```diff
@@ -46,4 +46,6 @@
 
 @property (nullable, copy) NSString<Optional> *currencyCode;
 
+@property (nonnull, copy) NSString  *positionDescription;
```
jsom: Is this always `nonnull`?

guillaumedebavelaere: It is set to empty string in the server by default:
`String description = ""`
It could be null if the broker returns null but in that case we should return the symbol (It is the case for IB). 
I could change it to nullable to be sure.

jsom: Yeah let's change it to nullable for now - I feel like some broker implementations won't have it.

PR #321, opened 2017-12-12
Gd/add launch transactions/153329272

guillaumedebavelaere:


TradeItIosTicketSDK2/TradeItLauncher.swift
```diff
@@ -332,4 +332,23 @@ protocol OAuthCompletionListener {
...
+        forLinkedBrokerAccount linkedBrokerAccount: TradeItLinkedBrokerAccount
+        ) {
+        deviceManager.authenticateUserWithTouchId(
+            onSuccess: {
+                let navController = self.viewControllerProvider.provideNavigationController(withRootViewStoryboardId: .transactionsView)
+
+                guard let transactionsViewController = navController.viewControllers.last as? TradeItTransactionsViewController else { return }
+                transactionsViewController.linkedBrokerAccount = linkedBrokerAccount
+                transactionsViewController.enableThemeOnLoad = false
+                transactionsViewController.view.backgroundColor = UIColor.tradeItlightGreyHeaderBackgroundColor
```
jsom: How come the theme is disabled?

guillaumedebavelaere: Enabling the theme would set some specific styling on the headers/table that doesn't look like the mocks.

jsom: OK - we should address this later if the mocks were from us. We need to conform to a specific set of styles because it makes maintaining the project harder. If they wanted to change the way headers look, then we should do it throughout the SDK with the theme configurator. Next time we get a mock that doesn't conform, let's raise it up and talk about it.

PR #323, opened 2017-12-20
Gd/yahoo broker selection notification/153373407

guillaumedebavelaere:
>Yahoo wants us to add instrumentation hooks for them on the broker list. They want to get a notification with:
>- UI Element Tapped - Label for row is probably fine (so Fidelity, etc)
>- Screen name
>- Is it a featured or a standard broker list click
>- We should also add a notification on the welcome broker screen page in case they are selecting Fidelity there

### DIFF
--- a/TradeItIosTicketSDK2/TradeItViewController.swift
+++ b/TradeItIosTicketSDK2/TradeItViewController.swift
@@ -29,42 +29,4 @@ class TradeItViewController: CloseableViewController {
             containerView.bottomAnchor.constraint(equalTo: self.view.bottomAnchor)
         ])
     }
-
-    func fireViewEventNotification(view: TradeItNotification.View, title: String? = nil) {
-        let title = title ?? "NO TITLE"
-        NotificationCenter.default.post(
-            name: TradeItNotification.Name.viewDidAppear,
-            object: nil,
-            userInfo: [
-                TradeItNotification.UserInfoKey.view.rawValue: view.rawValue,
-                TradeItNotification.UserInfoKey.viewTitle.rawValue: title
-            ]
-        )
-    }
-
-    func fireButtonTapEventNotification(view: TradeItNotification.View, button: TradeItNotification.Button) {
-        NotificationCenter.default.post(
-            name: TradeItNotification.Name.buttonTapped,
-            object: nil,
-            userInfo: [
-                TradeItNotification.UserInfoKey.view.rawValue: view.rawValue,
-                TradeItNotification.UserInfoKey.button.rawValue: button
-            ]
-        )
-    }
-
-    func fireLabelTappedEventNotification(view: TradeItNotification.View, title: String? = nil, labelText: String? = nil, label: TradeItNotification.Label) {
-        let title = title ?? "NO TITLE"
-        let labelText = labelText ?? "NO LABEL"
-        NotificationCenter.default.post(
-            name: TradeItNotification.Name.labelTapped,
-            object: nil,
-            userInfo: [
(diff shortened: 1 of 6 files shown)


guillaumedebavelaere:
I added the notification to the WelcomeViewController but I don't see where Yahoo is using it?

TradeItIosTicketSDK2/TradeItYahooBrokerSelectionViewController.swift
```diff
@@ -91,23 +91,34 @@ class TradeItYahooBrokerSelectionViewController: TradeItYahooViewController, UIT
...
     private func getBroker(atIndexPath indexPath: IndexPath) -> TradeItBroker {
-        if !self.featuredBrokers.isEmpty && indexPath.section == 0 {
+        if self.isFeaturedBrokerSelected(indexPath: indexPath) {
             return self.featuredBrokers[indexPath.row]
         } else {
             return self.brokers[indexPath.row]
         }
     }
 
+    private func isFeaturedBrokerSelected(indexPath: IndexPath) -> Bool {
```
jsom: Maybe rename to `isFeaturedCell(indexPath: IndexPath)` otherwise it could be confusing if someone selects a Featured Broker in the non-featured list.

TradeItIosTicketSDK2/TradeItYahooBrokerSelectionViewController.swift
```diff
@@ -91,23 +91,34 @@ class TradeItYahooBrokerSelectionViewController: TradeItYahooViewController, UIT
...
     // MARK: UITableViewDelegate
 
     func tableView(_ tableView: UITableView, didSelectRowAt indexPath: IndexPath) {
+        let isFeaturedBrokerSelected = self.isFeaturedBrokerSelected(indexPath: indexPath)
         let selectedBroker = self.getBroker(atIndexPath: indexPath)
+        self.fireLabelTappedEventNotification(
+            view: TradeItNotification.View.selectBroker,
+            title: self.title,
+            labelText: selectedBroker.brokerShortName,
+            label: isFeaturedBrokerSelected ? TradeItNotification.Label.featuredBroker : TradeItNotification.Label.broker
```
jsom: The use of `label` feels too broad. Are they actually tapping a label? Aren't they cells or buttons? We already have a button event notifier. Maybe these should be called `cellType` and `cellText`? That will make more sense if we end up adding this to all tables.

Thoughts @mitochondrion ?

TradeItIosTicketSDK2/TradeItViewController.swift
```diff
@@ -29,4 +29,42 @@ class TradeItViewController: CloseableViewController {
             containerView.bottomAnchor.constraint(equalTo: self.view.bottomAnchor)
         ])
     }
+
+    func fireViewEventNotification(view: TradeItNotification.View, title: String? = nil) {
```
jsom: If Y! isn't using the Welcome screen then we should skip adding it there for now. We can add it in easy enough later on but I don't want to add it as a feature to all of the TradeIt screens if it's still being figured out.

TradeItIosTicketSDK2/TradeItYahooBrokerSelectionViewController.swift
guillaumedebavelaere: I changed it to label because the featured broker in the welcome screen isn't a row. I can put it back to rowType/rowLabel and didSelectRow/onDidSelectRowNotification, since we remove it from the Welcome screen? (https://github.com/tradingticket/TradeItIosTicketSDK2/pull/323/commits/c176646271a6a8ddebec87e95a41b0b8092d70c3)

PR #324, opened 2017-12-20
[#153761760] Fix alerts when relinking on orders screen

guillaumedebavelaere:


jsom:
@guillaumedebavelaere this looks fine but can you describe why the alerts weren't showing because of this?

guillaumedebavelaere:
The `viewWillAppear` was called just before the oAuth completion (During relinking we see the screen from where you triggered the relinking), it was reloading the orders and an alert was triggered but never shown, so the `alreadyPresentingAlert ` variable was set to true and prevent the next alert to show up. https://github.com/tradingticket/TradeItIosTicketSDK2/blob/develop/TradeItIosTicketSDK2/TradeItAlertManager.swift#L218

PR #326, opened 2018-01-18
[#153690598] Fix Crashing when account is unlinked

guillaumedebavelaere:


TradeItIosTicketSDK2/TradeItOrdersViewController.swift
```diff
@@ -15,7 +15,8 @@ class TradeItOrdersViewController: TradeItViewController, TradeItOrdersTableDele
         super.viewDidLoad()
         
         guard let linkedBrokerAccount = self.linkedBrokerAccount else {
-            preconditionFailure("TradeItIosTicketSDK ERROR: TradeItOrdersViewController loaded without setting linkedBrokerAccount.")
+            print("TradeItIosTicketSDK ERROR: TradeItOrdersViewController loaded without setting linkedBrokerAccount.")
```
guillaumedebavelaere: Do we want to change `ERROR` to `WARNING`?

jsom:
This looks fine to me. @mitochondrion I think you introduced the use of `preconditionFailure` for this so does the change make sense for you?

PR #329, opened 2018-04-10
Branding to Link, Ticket, and Preview screens

oliver-was-here:
https://www.pivotaltracker.com/story/show/155711216

PR #335, opened 2018-04-23
Make branding optional

oliver-was-here:
Story: https://www.pivotaltracker.com/story/show/156885801

Yahoo is not prepared for branding in their SDK. This PR makes the necessary changes to the shared views in order to disable broker branding for the Yahoo implementation.

See most recent commit for relevant changes 03f6030


oliver-was-here:
Merging into #329 (review) to avoid conflicts that would ensue from rebasing #329 (review) onto the most up to date develop. `git rerere` was not doing enough for us here.

PR #338, opened 2018-05-02
Make UiConfig (Branding) optional at the SDK level

oliver-was-here:
Story: https://www.pivotaltracker.com/story/show/156885801

1. 8f4130e Allows for the configuration of UiConfig to enable/disable branding across all calls to the service by returning an empty array of configs without making a network request. **This solution makes the assumption that no partner integration calls __both__ `YahooLauncher` __and_ `TradeItLauncher`.
   - The changes to `ExampleViewController` should only be necessary in order to change `TradeItSDK.uiConfigService` on the fly after the application has been initialized (as our example app makes use of __both__ `YahooLauncher` and `TradeItLauncher`.
   - If Y! uses `TradeItSDK.tradeItLauncher` this PR may be inadequate. 
1. 13f31ff Removes the branding enable/disable on a per-cell basis. We may want this later but it serves no purpose for the time being. Should we go down this road we should allow for this configuration consistently across all branded cells.

### DIFF
--- a/ExampleApp/ExampleViewController.swift
+++ b/ExampleApp/ExampleViewController.swift
@@ -13,12 +13,14 @@ class Action {
     init(
         label: String,
         action: @escaping () -> Void,
-        oAuthCallbackUrl: String = "tradeItExampleScheme://completeOAuth"
+        oAuthCallbackUrl: String = "tradeItExampleScheme://completeOAuth",
+        isUiConfigServiceEnabled: Bool = true
     ) {
         self.label = label
         self.action = {
             TradeItSDK.oAuthCallbackUrl = URL(string: oAuthCallbackUrl)!
             action()
+            TradeItSDK.uiConfigService.isEnabled = isUiConfigServiceEnabled
         }
     }
 }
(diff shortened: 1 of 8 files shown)


oliver-was-here:
@jsom I'm not sure if I have enough context on this. The change only affects the `UiConfigService` and therein should only act to make the service behave in a NoOp fashion.

Poking around it doesn't seem that the `getBrokerList` depends on this config service in any way.

If the question is, does the featured broker logo still render on the broker selection screen the answer is no. However, as far as I can tell this was the behavior prior to the changes made in this PR -- as there seem to be no usages of the `logos` object which is returned by `getBrokerList`.

jsom:
Ah sorry, there was some legacy stuff that probably got removed before you started working on this track. Originally for featured brokers we pulled the logos from the `getBrokerList` endpoint. Then the broker branding push came through and so we made the more generalized `uiConfig` endpoint. Then we decided to keep the `uiConfig` changes off of Yahoo screens until we sorted everything out on the product side. So to summarise:
* TradeIt broker branding and featured brokers should get data from the `uiConfig` endpoint
* Yahoo featured brokers should get data from the `getBrokerList` endpoint which is the original behaviour. I'll dig up the old code.

jsom:
@oliver-was-here https://github.com/tradingticket/TradeItIosTicketSDK2/commit/207c7f0caa40c48ddf635a3fc59e42a3738ffba3#diff-9f36831739fd3ee0c115c11cc2a35fdb

PR #339, opened 2018-05-02
WIP - Yahoo crypto

jsom:


TradeItIosTicketSDK2/TradeItCryptoOrder.swift
```diff
@@ -0,0 +1,250 @@
...
+                onSuccess(
+                    result,
+                    self.generatePlaceOrderCallback(
+                        previewOrderResult: result
+                    )
+                )
+        }, onFailure: { error in
+            linkedBrokerAccount.linkedBroker?.error = error
+            onFailure(error)
+        }
```
jsom: Fix indentation

TradeItIosTicketSDK2/TradeItYahooCryptoTradingUIFlow.swift
```diff
@@ -0,0 +1,142 @@
+import UIKit
+
+class TradeItYahooCryptoTradingUIFlow:
```
jsom: This is a bunch of copy-pasta. Can probably create a protocol here.

TradeItIosTicketSDK2/BrandedTableViewCell.swift
```diff
@@ -0,0 +1,33 @@
+import UIKit
+
+class BrandedTableViewCell: UITableViewCell {
+    @IBOutlet weak var logoTitle: UIImageView!
+    @IBOutlet weak var logoWidthConstraint: NSLayoutConstraint!
+    @IBOutlet weak var altTitle: UILabel!
+
+    var adjustedLogoHeight: Double { return 14.0 }
```
falnatsheh: In this case is there benefits of having it to be a read-only property vs using `let` ?

PR #342, opened 2018-05-08
Version: 2.0.7

blanksblanks:
- [x] Make members of swift 4 classes accessible to obj-c
- [x] Fix example obj-c project
- [x] Point our app at the local pod to confirm all is well


TradeItIosTicketSDK2/TradeItYahooLauncher.swift
```diff
@@ -96,7 +96,7 @@ import SafariServices
         }
     }
 
-    public func launchTrading(
+    @objc public func launchTrading(
         fromViewController viewController: UIViewController,
         withOrder order: TradeItOrder,
         onViewPortfolioTappedHandler: @escaping OnViewPortfolioTappedHandler
```
blanksblanks: Verify this hook still works from obj-c as we use this in the Finance app.
Perhaps consider adding an invocation of this to the obj-c example?

The `typealias` for the handler can't be annotated with `obj-c` but passing in the long-form signature should work.

TradeItIosTicketSDK2/TradeItYahooLauncher.swift
```diff
@@ -53,7 +53,7 @@ import SafariServices
         )
     }
 
-    public func handleOAuthCallback(
+    @objc public func handleOAuthCallback(
         onTopmostViewController topMostViewController: UIViewController,
         oAuthCallbackUrl: URL,
         onOAuthCompletionSuccessHandler: OnOAuthCompletionSuccessHandler? = nil
```
blanksblanks: Verify this works. Same situation/comment as `onViewPortfolioTappedHandler` basically.

TradeItIosTicketSDK2/Info.plist
```diff
@@ -1,5 +1,5 @@
 <?xml version="1.0" encoding="UTF-8"?>
-<!DOCTYPE plist PUBLIC "-//Apple//DTD PLIST 1.0//EN" "http://www.apple.com/DTDs/PropertyList-1.0.dtd">
+<!DOCTYPE plist @objc public "-//Apple//DTD PLIST 1.0//EN" "http://www.apple.com/DTDs/PropertyList-1.0.dtd">
```
jsom: Need to remove

TradeItIosTicketSDK2/AdService.swift
```diff
@@ -61,7 +61,7 @@ import UIKit
```
blanksblanks: @jsom Question about the comment that was there before on `labelFor` in the above enums about obj-c backwards compatibility.

When Swift enums are exposed to obj-c, only their raw types are visible, right? So if obj-c cannot see this `labelFor` method, what's the comment re:compatibility about?

jsom: Obj-c does not support Swift non-integer enums AFAIK. If we didn't need to support obj-c then we could make this a `public enum TradeItAdPosition: String` and get rid of the `labelFor` - I haven't looked at this for a while though so maybe something changed.

TradeItIosTicketSDK2/Info.plist
blanksblanks: good catch. thanks!

TradeItIosTicketSDK2/TradeItYahooLauncher.swift
jsom: @blanksblanks did you verify this?

PR #349, opened 2018-05-25
Allow partners to provide streaming market data for the Trading Ticket

oliver-was-here:
https://www.pivotaltracker.com/story/show/157706133


ExampleApp/AppDelegate.swift
```diff
@@ -234,3 +234,44 @@ class DummyMarketDataService: MarketDataService {
...
+    var onUpdate: ((TradeItQuote) -> Void)?
+    
+    @objc func updateQuote() {
+        let price = Double("\(arc4random_uniform(8000)).\(arc4random_uniform(99))")
+        
+        let quote = TradeItQuote()
+        quote.companyName = "Bitcoin"
+        quote.lastPrice = NSNumber(value: lastPrice)
+        quote.bidPrice = NSNumber(value: price!)
+        quote.askPrice = NSNumber(value: price!)
```
jsom: Avoid force unwrapping even in dummy/example code. Partners might copy paste it.

Podfile.lock
```diff
@@ -43,4 +34,4 @@ SPEC CHECKSUMS:
 
 PODFILE CHECKSUM: 8de6c042a9db9777f64178e4995be9fd85814f2f
 
-COCOAPODS: 1.5.0
+COCOAPODS: 1.4.0
```
jsom: Don't change this. Also this should be against the `swift-4` branch.

TradeItIosTicketSDK2/TradeItSDK.swift
```diff
@@ -82,6 +82,17 @@ import UIKit
             return self._brokerCenterService!
         }
     }
+    
+    internal static var _streamingMarketDataService: StreamingMarketDataService?
+    public static var streamingMarketDataService: StreamingMarketDataService? {
+        get {
+            return self._streamingMarketDataService
```
jsom: If there is no pre-condition check needed then you don't need the getter/setter.

TradeItIosTicketSDK2/TradeItTradingTicketViewController.swift
```diff
@@ -86,6 +94,16 @@ class TradeItTradingTicketViewController: TradeItViewController, UITableViewData
             self.initializeTicket()
         } else {
             self.reloadTicketRows()
+            if let streamingMarketDataService = streamingMarketDataService,
+                let symbol = self.order.symbol {
+                streamingMarketDataService.startUpdatingQuote(forSymbol: symbol, onUpdate: updateQuotePrice, onFailure: { self.clearMarketData() })
```
jsom: `onFailure: self.clearMarketData`?

TradeItIosTicketSDK2/TradeItTradingTicketViewController.swift
```diff
@@ -86,6 +94,16 @@ class TradeItTradingTicketViewController: TradeItViewController, UITableViewData
...
+            }
+        }
+    }
```
jsom: Does this handle if the symbol is changed from the ticket?

TradeItIosTicketSDK2/TradeItTradingTicketViewController.swift
```diff
@@ -25,6 +26,13 @@ class TradeItTradingTicketViewController: TradeItViewController, UITableViewData
     private var equityOrderCapabilities: TradeItInstrumentOrderCapabilities?
 
     private var selectedAccountChanged: Bool = true
+    
+    private lazy var updateQuotePrice: (TradeItQuote) -> Void = { quote in
+        self.quote = quote
+        self.order.quoteLastPrice = TradeItQuotePresenter.numberToDecimalNumber(quote.lastPrice)
+        self.reload(row: .marketPrice)
+        self.reload(row: .estimatedCost)
```
jsom: Is there a way to update the price label without updating the entire row? The fading animation of the whole row is a little distracting.

ExampleApp/AppDelegate.swift
```diff
@@ -234,3 +234,44 @@ class DummyMarketDataService: MarketDataService {
         onFailure(error)
     }
 }
+
+// Only implement this protocol if you need to stream your own market data
+class DummyStreamingMarketDataService: StreamingMarketDataService {
```
jsom: Is this on by default now in the SDK? If not, is there an easy way to test it from the ExampleApp? Maybe we can add a `Toggle Dummy streaming` button to the example app.

ExampleApp/AppDelegate.swift
```diff
@@ -234,3 +234,44 @@ class DummyMarketDataService: MarketDataService {
...
+    var timer: Timer?
+    var lastPrice = 2000.0
+    var onUpdate: ((TradeItQuote) -> Void)?
+    
+    @objc func updateQuote() {
+        let price = Double("\(arc4random_uniform(8000)).\(arc4random_uniform(99))")
```
jsom: Can we make this just do a random walk instead of totally random numbers?

TradeItIosTicketSDK2/TradeItTradingTicketViewController.swift
falnatsheh: It does but the logic was: 
1. Start streaming service on view did appear
2. Start streaming service on Symbol change. 

So if the user changed the symbol and then went back to the view then 1 and 2 will be called. I added a check to prevent double `start` calls, I choose a simple approach by tracking `selectedSymbolChanged`.

TradeItIosTicketSDK2/TradeItTradingTicketViewController.swift
falnatsheh: I added optional parameter `reload(row: TicketRow, animation: UITableViewRowAnimation = .automatic)`. So if you call it with `.none` like in this example `self.reload(row: .marketPrice, animation: .none)` you will not see the fading animation. It looked good when the price only changed will try to see how it will look if other part of the row changed.

TradeItIosTicketSDK2/TradeItTradingTicketViewController.swift
```diff
@@ -242,6 +248,7 @@ class TradeItTradingTicketViewController: TradeItViewController, UITableViewData
         self.order.symbol = selectedSymbol
         self.clearMarketData()
         self.updateMarketData()
+        self.selectedSymbolChanged = true
```
oliver-was-here: maybe we make this `self.selectedSymbolChanged = self.order.symbol != selectedSymbol`? IMO it fits better definition wise.

This controller flow is a little confusing as well -- not sure we can do much about it right now, but it requires knowing a thing or two about when different actions are called

TradeItIosTicketSDK2/TradeItTradingTicketViewController.swift
```diff
@@ -221,6 +238,7 @@ class TradeItTradingTicketViewController: TradeItViewController, UITableViewData
         _ symbolSearchViewController: TradeItSymbolSearchViewController,
         didSelectSymbol selectedSymbol: String
     ) {
+        self.selectedSymbolChanged = true
```
oliver-was-here: maybe we make this `self.selectedSymbolChanged = self.order.symbol != selectedSymbol`? IMO it fits better definition wise.

This controller flow is a little confusing as well -- not sure we can do much about it right now, but it requires knowing a thing or two about when different actions are called

jsom: Do we reset this to `false` after we use it?

jsom: Agreed that this is getting confusing.

ExampleApp/AppDelegate.swift
```diff
@@ -236,11 +236,12 @@ class DummyMarketDataService: MarketDataService {
 }
 
 // Only implement this protocol if you need to stream your own market data
-class DummyStreamingMarketDataService: StreamingMarketDataService {
+public class DummyStreamingMarketDataService: StreamingMarketDataService {
```
jsom: Why public? We don't want partners to start using this market data service ;) Or do you need to make it public for `ExampleApp`?

falnatsheh: I thought it needed to be public for `ExampleViewController` toggle button 😅. Reverting it to default since it's fine without it.

Podfile.lock
```diff
@@ -24,15 +24,6 @@ DEPENDENCIES:
...
 
-SPEC REPOS:
-  https://github.com/CocoaPods/Specs.git:
-    - BEMCheckBox
-    - JSONModel
-    - MBProgressHUD
-    - Nimble
-    - PromiseKit
-    - Quick
-
```
jsom: Why did this change?

TradeItIosTicketSDK2.xcodeproj/project.pbxproj
```diff
@@ -2618,6 +2683,21 @@
...
+			inputPaths = (
+			);
+			name = "[CP] Copy Pods Resources";
+			outputPaths = (
+			);
+			runOnlyForDeploymentPostprocessing = 0;
+			shellPath = /bin/sh;
+			shellScript = "\"${SRCROOT}/Pods/Target Support Files/Pods-ExampleAppUITests/Pods-ExampleAppUITests-resources.sh\"\n";
+			showEnvVarsInLog = 0;
+		};
```
jsom: Why did this change? Did you run `pod install` with the latest version of cocoapods?

TradeItIosTicketSDK2/TradeItTradingTicketViewController.swift
```diff
@@ -81,12 +89,20 @@ class TradeItTradingTicketViewController: TradeItViewController, UITableViewData
             )
             return
         }
-
+
```
jsom: @falnatsheh what's going on with your Xcode whitespace trimmer!

TradeItIosTicketSDK2.xcodeproj/project.pbxproj
oliver-was-here: Au contraire, an old version of cocopods I believe (`1.4.x`). This was run on the pair station. What's the move, re-run `pod install` with an up to date cocopods?

TradeItIosTicketSDK2/TradeItTradingTicketViewController.swift
falnatsheh: So this is inside a function to separate logical group.

Podfile.lock
falnatsheh: Yeah this might be an auto change (when running pod install) on the pairing machine. I'll reverted back.

TradeItIosTicketSDK2.xcodeproj/project.pbxproj
jsom: Yes please

TradeItIosTicketSDK2/TradeItTradingTicketViewController.swift
jsom: Yeah but these are just blank space characters

PR #351, opened 2018-05-31
Incorrect currency displayed on Portfolio Screen

oliver-was-here:
https://www.pivotaltracker.com/story/show/157233356

### DIFF
--- a/TradeItIosTicketSDK2/NumberFormatter.swift
+++ b/TradeItIosTicketSDK2/NumberFormatter.swift
@@ -2,7 +2,13 @@ class NumberFormatter: NSObject {
     private static let currencyFormatter = Foundation.NumberFormatter()
     private static let quantityFormatter = Foundation.NumberFormatter()
     
-    static func formatCurrency(_ number: NSNumber, minimumFractionDigits: Int = 2, maximumFractionDigits: Int = 2, displayVariance: Bool = false, currencyCode: String? = TradeItPresenter.DEFAULT_CURRENCY_CODE) -> String {
+    static func formatCurrency(
+        _ number: NSNumber,
+        minimumFractionDigits: Int = 2,
+        maximumFractionDigits: Int = 2,
+        displayVariance: Bool = false,
+        currencyCode: String? = TradeItPresenter.DEFAULT_CURRENCY_CODE
+    ) -> String {
         currencyFormatter.numberStyle = .currency
         currencyFormatter.currencyCode = currencyCode
         currencyFormatter.currencySymbol = overrideCurrencySymbol(forCurrencyCode: currencyCode)
--- a/TradeItIosTicketSDK2/NumberFormatter.swift
+++ b/TradeItIosTicketSDK2/NumberFormatter.swift
@@ -12,18 +12,17 @@ class NumberFormatter: NSObject {
         currencyFormatter.numberStyle = .currency
         currencyFormatter.currencyCode = currencyCode
         currencyFormatter.currencySymbol = overrideCurrencySymbol(forCurrencyCode: currencyCode)
-        if displayVariance {
-            currencyFormatter.positivePrefix = "+" + currencyFormatter.currencySymbol
-            currencyFormatter.negativePrefix = "-" + currencyFormatter.currencySymbol
-        } else {
-            currencyFormatter.positivePrefix = nil
-            currencyFormatter.negativePrefix = nil
-        }
         currencyFormatter.minimumFractionDigits = minimumFractionDigits
         currencyFormatter.maximumFractionDigits = maximumFractionDigits
-        return currencyFormatter.string(from: number) ?? TradeItPresenter.MISSING_DATA_PLACEHOLDER
+        var formattedCurrency = currencyFormatter.string(from: number)
+        if displayVariance,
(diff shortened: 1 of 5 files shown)


TradeItIosTicketSDK2/NumberFormatter.swift
```diff
@@ -2,22 +2,27 @@ class NumberFormatter: NSObject {
...
-            currencyFormatter.negativePrefix = nil
-        }
         currencyFormatter.minimumFractionDigits = minimumFractionDigits
         currencyFormatter.maximumFractionDigits = maximumFractionDigits
-        return currencyFormatter.string(from: number) ?? TradeItPresenter.MISSING_DATA_PLACEHOLDER
+        var formattedCurrency = currencyFormatter.string(from: number)
+        if displayVariance,
+            number.doubleValue > 0.0,
+            let unwrappedFormattedCurrency = formattedCurrency {
+            formattedCurrency = "+\(unwrappedFormattedCurrency)"
```
oliver-was-here: `NumberFormatter#string()` auto prepends `-` out of the box. Presumably we _always_ want to display the `-` sign for negative numbers -- so we need only account for `+`. 

`NumberFormatter` provides some config options (`negativePrefix`, `negativeFormat`, ...), however, they all come with quirks (not respect locale formatting for currencies, and whatever else Apple shoves into their black box.)

TradeItIosTicketSDK2/TradeItPortfolioBalanceEquityPresenter.swift
```diff
@@ -44,7 +44,7 @@ class TradeItPortfolioBalanceEquityPresenter {
         var dayReturnString = ""
 
         if let dayAbsoluteReturn = self.balance?.dayAbsoluteReturn {
-            dayReturnString = NumberFormatter.formatCurrency(dayAbsoluteReturn, displayVariance: true, currencyCode: balance?.accountBaseCurrency)
+            dayReturnString = NumberFormatter.formatCurrency(dayAbsoluteReturn, displayVariance: true)
```
jsom: Why remove this? If the balance returned is not USD then this won't work.

TradeItIosTicketSDK2/TradeItPortfolioBalanceEquityPresenter.swift
```diff
@@ -44,7 +44,7 @@ class TradeItPortfolioBalanceEquityPresenter {
         var dayReturnString = ""
 
         if let dayAbsoluteReturn = self.balance?.dayAbsoluteReturn {
-            dayReturnString = NumberFormatter.formatCurrency(dayAbsoluteReturn, displayVariance: true, currencyCode: balance?.accountBaseCurrency)
```
oliver-was-here: If `currencyCode` is `nil`, `currencySymbol` defaults to the symbol dictated by `locale`

jsom: I don't think we want to default to `locale` though - if I'm set to the Thai locale and my balance returns in AUD then I want to see AUD

TradeItIosTicketSDK2/TradeItPortfolioBalanceEquityPresenter.swift
oliver-was-here: ah shizen, good catch. had mean to circle back with @falnatsheh and think of the best way to handle this.

Options that first come to mind are:
1. `dayReturnString = NumberFormatter.formatCurrency(dayAbsoluteReturn, displayVariance: true, currencyCode: balance?.accountBaseCurrency ?? TradeItPresenter.DEFAULT_CURRENCY_CODE` 
1. Swap out the following code with `currencyFormatter.currencyCode = currencyCode ?? TradeItPresenter.DEFAULT_CURRENCY_CODE`
https://github.com/tradingticket/TradeItIosTicketSDK2/blob/be032b9ebaf53871795758433ba8a032d7cafe28/TradeItIosTicketSDK2/NumberFormatter.swift#L13

I think I'd lean toward option 2.

TradeItIosTicketSDK2/TradeItPortfolioBalanceEquityPresenter.swift
oliver-was-here: Whoops, this comment was intended for the dif on `TradeItPortfolioAccountsTableViewManager.swift` `L110`

TradeItIosTicketSDK2/TradeItPortfolioBalanceEquityPresenter.swift
jsom: Maybe the right thing to do is to ignore the value if we can't determine a currency code? Just return `nil`.

oliver-was-here: good point -- take a look at https://github.com/tradingticket/TradeItIosTicketSDK2/pull/351/commits/a2434727fbda466a1386e9362185a8fcadbac142 and lmk if I'm Swift-ing right

oliver-was-here:
@jsom this is testable, instructions to test are in the [story comments](https://www.pivotaltracker.com/story/show/157233356/comments/190170308)

PR #352, opened 2018-06-03
Disable highlighting when TableViewCell or the Radio Button within th…

oliver-was-here:
Story: https://www.pivotaltracker.com/story/show/157826780

Solution proposed in [this](https://stackoverflow.com/a/191245/3837185) SO post


jsom:
@oliver-was-here there should be a way to do this via the storyboard which is what we've been doing for the rest of the SDK

oliver-was-here:
@jsom Ah, I'd poked around for it but wasn't able to find it. Might have had the wrong cell selected. We also have 4 other usages of `cell.selectionStyle = .none` in our codebase. Not sure if we want to address those now or later/never

https://github.com/tradingticket/TradeItIosTicketSDK2/blob/3c11bfc21273f88a495923a50b0cadab5ac96764/TradeItIosTicketSDK2/TradeItTradingTicketViewController.swift#L446

PR #353, opened 2018-06-04
Make the low and high properties in the TradeItQuote response optional

oliver-was-here:
https://www.pivotaltracker.com/story/show/158016177

A more lenient deserialization policy makes custom implementation of the `StreamingMarketDataService` very straightforward for partners who wish to use `JSONModel` (see https://github.com/tradingticket/TradeItIosTicketSDK2/pull/349 for more context).

For example, all the partner need provide is the following mapping information (from their backing websocket response to the property names in `TradeItQuote`)
```@implementation TradeItQuote

+ (JSONKeyMapper *)keyMapper
{
    return [[JSONKeyMapper alloc] initWithModelToJSONDictionary:@{
                                                                  @"symbol": @"partnerSymbolKey",
                                                                  @"lastPrice": @"partnerPriceKey"
                                                                  }];
}


@end
```


jsom:
Are we using these properties anywhere in the SDK?

oliver-was-here:
@jsom looks like we access them in a nil-safe manner here:
https://github.com/tradingticket/TradeItIosTicketSDK2/blob/3c11bfc21273f88a495923a50b0cadab5ac96764/TradeItIosTicketSDK2/TradeItPortfolioPositionPresenter.swift#L47-L49


and here:
https://github.com/tradingticket/TradeItIosTicketSDK2/blob/3c11bfc21273f88a495923a50b0cadab5ac96764/TradeItIosTicketSDK2/TradeItPortfolioPositionPresenter.swift#L28-L30


aside from that we have the following method which is unused that provides data via a `Dictionary` (is this old code?)
https://github.com/tradingticket/TradeItIosTicketSDK2/blob/3c11bfc21273f88a495923a50b0cadab5ac96764/TradeItIosEmsApi/TradeItQuote.m#L22-L23

jsom:
OK. Should be fine. Properties are already `nullable` and this is just allowing JSONModel to treat them as such. Shouldn't impact anyone using the `TradeItQuote` object directly.

PR #358, opened 2018-06-08
Static lib

falnatsheh:
To build a static lib the developer has to specify `use_modular_headers!` in their pod file. If our SDK is used in Objective-C code the user need to replace `#import <TradeItIosTicketSDK2/TradeItIosTicketSDK2-Swift.h>` with `@import TradeItIosTicketSDK2.Swift;`. 

You could test this branch by adding the following to `Podfile`: `pod 'TradeItIosTicketSDK2', :git => 'https://github.com/tradingticket/TradeItIosTicketSDK2.git', :branch => 'static-lib'`

**Note:** `@import TradeItIosTicketSDK2.Swift;` also works if the developer has `use_frameworks!` in their pod file.

### DIFF
--- a/Cartfile
+++ b/Cartfile
@@ -1,4 +1,4 @@
-github "Boris-Em/BEMCheckBox" "1.4.1"
-github "JSONModel/JSONModel" "1.7.0"
-github "jdg/MBProgressHUD" "1.1.0"
-github "mxcl/PromiseKit" "6.2.8"
+github "Boris-Em/BEMCheckBox" ~> 1.4.1
+github "JSONModel/JSONModel" ~> 1.7.0
+github "jdg/MBProgressHUD" ~> 1.0.0
+github "mxcl/PromiseKit" ~> 6.2.8
--- a/Cartfile
+++ b/Cartfile
@@ -1,4 +1,4 @@
-github "mxcl/PromiseKit" ~> 4.4
-github "jdg/MBProgressHUD" ~> 1.0.0
-github "JSONModel/JSONModel" "master"
-github "Boris-Em/BEMCheckBox" ~> 1.4.1
+github "Boris-Em/BEMCheckBox" "1.4.1"
+github "JSONModel/JSONModel" "1.7.0"
+github "jdg/MBProgressHUD" "1.1.0"
+github "mxcl/PromiseKit" "6.2.8"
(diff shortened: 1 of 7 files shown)


TradeItIosEmsApi/TradeItConnector.m
```diff
@@ -12,7 +12,11 @@
 #ifdef CARTHAGE
 #import <TradeItIosTicketSDK2Carthage/TradeItIosTicketSDK2Carthage-Swift.h>
 #else
-#import <TradeItIosTicketSDK2/TradeItIosTicketSDK2-Swift.h>
```
falnatsheh: We can't use `@import TradeItIosTicketSDK2.Swift;` since we are inside the module. This is a workaround for the way Cocopod build static libs and we might be able to remove in the future if it was fixed on their side.

jsom:
@fandyfyf wanna give this a try?

@falnatsheh has this been tested with a Carthage app too?

falnatsheh:
@jsom Not yet, will tackle that next.

falnatsheh:
Updated Cartage dependancies and test it locally with Swift/ObjC projects ✅

Cartfile
```diff
@@ -1,4 +1,4 @@
-github "mxcl/PromiseKit" ~> 4.4
-github "jdg/MBProgressHUD" ~> 1.0.0
-github "JSONModel/JSONModel" "master"
-github "Boris-Em/BEMCheckBox" ~> 1.4.1
+github "Boris-Em/BEMCheckBox" "1.4.1"
+github "JSONModel/JSONModel" "1.7.0"
+github "jdg/MBProgressHUD" "1.1.0"
+github "mxcl/PromiseKit" "6.2.8"
```
jsom: @falnatsheh any reason to pin to exact versions here?

falnatsheh: The reason I used these specific values because it matches with what we used with CocoaPod. This way the dependencies stay consistent between CocoaPod and Carthage.

jsom: Hmm - looks like our Cocoapods is set to `~>` which will incorporate updates as per:
```
'~> 0.1.2' Version 0.1.2 and the versions up to 0.2, not including 0.2 and higher
'~> 0.1' Version 0.1 and the versions up to 1.0, not including 1.0 and higher
'~> 0' Version 0 and higher, this is basically the same as not having it.
```

falnatsheh: Thanks for pointing that out, it looks like Carthage has similar operator (from their [documentations](https://github.com/Carthage/Carthage/blob/master/Documentation/Artifacts.md#version-requirement)): `~> 1.0 for “compatible with version 1.0”`. So this could be updated to: 

```
github "Boris-Em/BEMCheckBox" ~> 1.4.1
github "JSONModel/JSONModel" ~> 1.7.0
github "jdg/MBProgressHUD" ~> 1.0.0
github "mxcl/PromiseKit" ~> 6.2.8
```

PR #362, opened 2018-06-13
Order Quantity Type + Amount Change Crash Fix

falnatsheh:
- https://www.pivotaltracker.com/story/show/158305029
- https://www.pivotaltracker.com/story/show/158329465
- https://www.pivotaltracker.com/story/show/157754041


TradeItIosTicketSDK2/TradeItYahooEquityTradingTicketViewController.swift
```diff
@@ -416,8 +416,25 @@ class TradeItYahooEquityTradingTicketViewController: TradeItYahooViewController,
...
+                    code: TradeItErrorCode.systemError
+                )
+                self.alertManager.showError(errorResult, onViewController: self)
+                return cell
+            }
+            
+            self.order.quantityType = supportedOrderQuantityTypes.contains(self.order.quantityType) ? self.order.quantityType : supportedOrderQuantityTypes[0]
             let quantitySymbol = self.order.quantityType == OrderQuantityType.shares ? "Shares" : self.order.linkedBrokerAccount?.accountBaseCurrency
+            
+            let cell = cell as? TradeItNumericToggleInputCell
```
falnatsheh: I moved this down so I could use the previous none optional `cell` as return if an error occurred.

TradeItIosTicketSDK2/TradeItYahooEquityTradingTicketViewController.swift
```diff
@@ -416,8 +416,25 @@ class TradeItYahooEquityTradingTicketViewController: TradeItYahooViewController,
...
+            let supportedOrderQuantityTypes = orderCapabilities.supportedOrderQuantityTypes(
+                forAction: self.order.action,
+                priceType: self.order.type
+            )
+            
+            if supportedOrderQuantityTypes.isEmpty {
+                let errorResult = TradeItErrorResult(
+                    title: "Order Quantity Type Error",
+                    message: "No valid quantity type found for this action",
+                    code: TradeItErrorCode.systemError
```
falnatsheh: Not sure what error code this fits under, this seems the closest but let me know if you I need to change it.

TradeItIosTicketSDK2/TradeItYahooEquityTradingTicketViewController.swift
```diff
@@ -416,12 +416,29 @@ class TradeItYahooEquityTradingTicketViewController: TradeItYahooViewController,
...
+            
+            self.order.quantityType = supportedOrderQuantityTypes.contains(self.order.quantityType) ? self.order.quantityType : supportedOrderQuantityTypes[0]
             let quantitySymbol = self.order.quantityType == OrderQuantityType.shares ? "Shares" : self.order.linkedBrokerAccount?.accountBaseCurrency
+            
+            let cell = cell as? TradeItNumericToggleInputCell
             cell?.configure(
                 onValueUpdated: { newValue in
                     self.order.quantity = newValue
-                    self.reload(row: .estimatedChange)
+                    self.updateEstimatedChangeRow()
```
falnatsheh: Below are the changes to fix the crash when changing the estimated change row.

falnatsheh:
@jsom if you have time when you do acceptance for #158329465, can you try to see if you could better understand TableView inconsistent state that causes the error. So far what I think is that if you try to call `reloadRows` inside `tableView:cellForRowAt` sometimes the TableView gets into bad state, replacing `reloadRows` with updating the cell directly fixes the issue.

TradeItIosTicketSDK2/TradeItYahooEquityTradingTicketViewController.swift
```diff
@@ -416,12 +416,29 @@ class TradeItYahooEquityTradingTicketViewController: TradeItYahooViewController,
...
+                let errorResult = TradeItErrorResult(
+                    title: "Order Quantity Type Error",
+                    message: "No valid quantity type found for this action",
+                    code: TradeItErrorCode.systemError
+                )
+                self.alertManager.showError(errorResult, onViewController: self)
+                return cell
+            }
+            
+            self.order.quantityType = supportedOrderQuantityTypes.contains(self.order.quantityType) ? self.order.quantityType : supportedOrderQuantityTypes[0]
```
jsom: This code can still explode if someone accidentally removes the `isEmpty` check above. Better to use a guard?

```swift
guard let firstSupportedOrderQuantityType = supportedOrderQuantityTypes.first
    else {
        // show error
    }

self.order.quantityType = supportedOrderQuantityTypes.contains(self.order.quantityType) ? self.order.quantityType : firstSupportedOrderQuantityType
```

TradeItIosTicketSDK2/TradeItYahooEquityTradingTicketViewController.swift
```diff
@@ -529,6 +533,26 @@ class TradeItYahooEquityTradingTicketViewController: TradeItYahooViewController,
...
+    private func estimatedChangeText() -> String {
+        var estimateChangeText = "N/A"
+        
+        if let estimatedChange = order.estimatedChange() {
+            estimateChangeText = NumberFormatter.formatCurrency(
+                estimatedChange,
+                currencyCode: order.linkedBrokerAccount?.accountBaseCurrency
+            )
+        }
+        return estimateChangeText
```
jsom: I usually prefer a guard in cases like this to reduce nesting:
```swift
guard let estimatedChange = order.estimatedChange()
    else {
        return "N/A"
    }
return NumberFormatter.formatCurrency(
    estimatedChange,
    currencyCode: order.linkedBrokerAccount?.accountBaseCurrency
)
```

TradeItIosTicketSDK2/TradeItYahooEquityTradingTicketViewController.swift
falnatsheh: Yeah the more swift-ey way of doing it

TradeItIosTicketSDK2/EquityPreviewCellFactory.swift
```diff
@@ -72,6 +72,19 @@ class EquityPreviewCellFactory: PreviewCellFactory {
 
     // MARK: Private
 
+    private func formatQuantity(rawQuantityType: String, quantity: NSNumber) -> String{
```
falnatsheh: Ticket: https://www.pivotaltracker.com/story/show/157754041

Not very happy with the method but this is my first stab at it. Notes: 

- The `NumberFormatter.formatQuantity` defaults to 2, we could change/add "Quantity formatter" method that uses `orderCapabilities?.maxDecimalPlacesFor`.
- `orderDetails.orderQuantityType` is a string that in order to use it with `orderCapabilities?.maxDecimalPlacesFor` need to be convert to enum. If `EquityPreviewCellFactory.swift`  received the full `order` object or orderQuantityType enum the conversion will not be needed.

jsom: @falnatsheh yeah I think adding a new parameter to `NumberFormatter.formatQuantity` makes sense

jsom:
@falnatsheh there also seems to be a WIP commit at the top here. Can you update it with a good commit message if you want to keep it?

PR #369, opened 2018-06-25
[WIP] Meta POC

jsom:


ExampleApp/equityTicketConfig.json
```diff
@@ -0,0 +1,173 @@
...
+                "label": "Quantity",
+                "inputType": "decimal"
+                },
+                {
+                "apiParameter": "orderLimit",
+                "label": "Limit",
+                "inputType": "decimal",
+                "dependencies": [
+                {
+                                 "condition": "==",
```
oliver-was-here: `condition` seems like it would be better named as `operator` -- as I see it the entirety of the statement is the `condition`, not the type of operator being used.

ExampleApp/equityTicketConfig.json
```diff
@@ -0,0 +1,195 @@
...
+					"label": "Shares"
+				},
+				{
+					"value": "TOTAL_VALUE",
+					"label": "USD",
+					"dependencies": [
+						{
+							"condition": "==",
+							"value": "buy",
+							"criteria": "orderAction"
```
oliver-was-here: I'm not hot on `criteria`. `fieldId` or `fieldKey` seems nicer.

ExampleApp/equityTicketConfig.json
```diff
@@ -0,0 +1,195 @@
...
+			"inputType": "options",
+			"options": [
+				{
+					"value": "SHARES",
+					"label": "Shares"
+				},
+				{
+					"value": "TOTAL_VALUE",
+					"label": "USD",
+					"dependencies": [
```
oliver-was-here: 1. I'm not hot on `dependencies`. Was this inspired by JSON Schema? `prerequisites` seems to carry more meaning.
2. It is unclear to me how we are logically chaining together `dependencies`. My initial take was that these were `&&`’d together, however, as there are recurrent entries with the same `criteria` it would seem we are `||`’ing them. Perhaps we don’t need both operator types in the present moment, however, the cost of doing it later on could be quite high — and as I said it’s not obvious in every payload that these are `||`s to start with.

Perhaps we add reformat to the following:
```json
"dependencies": {
   "operator": "||"          // can be || or &&, ignore in the case of a 
   "conditions": [
       {
          "operator": "==",
			"value": "limit",
			"criteria": "orderPrice"
       }, 
       {
			"operator": "==",
			"value": "stopLimit",
			"criteria", "orderPrice"
       }
    ]
}
```

Thinking outward, it would also be possible to nest conditions by supplying `||` or `&&`  along with a list of `conditions` within a specific condition.

ExampleApp/equityTicketConfig.json
```diff
@@ -0,0 +1,195 @@
...
+						{
+							"condition": "==",
```
oliver-was-here: I'm not hot on `condition`. From a boolean logic perspective, this entire object is a condition, and this is the `operator`.

jsom: Yah agreed, this was taken from something thrown together by Serge - try not to focus too much on the naming for now and try to evaluate on:
1. How easy it would be to create new clients (for both us on different platforms and third parties creating custom tickets)
1. Does it cover all of the cases we currently have (reuse same VC for equity/crypto/etc)
1. Do we prefer this style of dependencies/prerequisites lists vs a tree structure

ExampleApp/equityTicketConfig.json
jsom: Yeah I had thought about this too but at that point do we make it totally generic and allow recursive conditions? Will that drastically complicate a new integration (maybe unnecessarily)?

oliver-was-here: If we only implement one layer deep comparisons for the time being the same general schema could be expanded upon later, whereas this current schema require a rehaul. The client code will still need to be updated, however, I don't see how this is any different than the fact we supply `condition` but won't be providing every possible condition (`>`, `<`, etc...) in V1.

falnatsheh: Related to the logical operators, this is what MongoDB do: 
```
   $and : [
        { $or : [ { price : 0.99 }, { price : 1.99 } ] },
        { $or : [ { sale : true }, { qty : { $lt : 20 } } ] }
    ]
```
And this is DB2:
`{"$or":[{"age":4},{"name":"Joe"}]}`

These kind of formats allows for a lot of flexibility but I think It might be overkill for our needs.

PR #374, opened 2018-07-19
Add LinkedBrokers with a session error to the DisplayableLinkedBrokers

jsom:
`authenticateAll` was using `getAllDisplayableLinkedBrokers` which was filtering out any linked broker without any accounts. This is a problem because when a broker is just linked and not yet authenticated there are 0 accounts. `authenticateAll` was then never authenticating it. I checked the other usages of `getAllDisplayableLinkedBrokers` and this seems to be a safe change.

### DIFF
--- a/TradeItIosTicketSDK2/TradeItLinkedBrokerManager.swift
+++ b/TradeItIosTicketSDK2/TradeItLinkedBrokerManager.swift
@@ -225,7 +225,7 @@ import PromiseKit
     }
     
     @objc public func getAllDisplayableLinkedBrokers() -> [TradeItLinkedBroker] {
-        return self.linkedBrokers.filter { $0.getEnabledAccounts().count > 0 || $0.isAccountLinkDelayedError}
+        return self.linkedBrokers.filter { $0.getEnabledAccounts().count > 0 || $0.error?.requiresAuthentication() == true }
     }
     
     @objc public func getAllActivationInProgressLinkedBrokers() -> [TradeItLinkedBroker] {


TradeItIosTicketSDK2/TradeItLinkedBrokerManager.swift
```diff
@@ -225,7 +225,7 @@ import PromiseKit
     }
     
     @objc public func getAllDisplayableLinkedBrokers() -> [TradeItLinkedBroker] {
-        return self.linkedBrokers.filter { $0.getEnabledAccounts().count > 0 || $0.isAccountLinkDelayedError}
+        return self.linkedBrokers.filter { $0.getEnabledAccounts().count > 0 || $0.isAccountLinkDelayedError || $0.error?.errorCode == TradeItErrorCode.sessionError }
```
guillaumedebavelaere: Maybe we can just use the `requiresAuthentication` method?
```
return self.linkedBrokers.filter { $0.getEnabledAccounts().count > 0 || $0.error?.requiresAuthentication() }
```

It contains in addition, the "brokerAccountError": https://github.com/tradingticket/TradeItIosTicketSDK2/blob/ed964cf9730a81a31f98f055a2ec341afb4dad89/TradeItIosTicketSDK2/TradeItErrorResult%2BErrorCodeHelpers.swift#L81

jsom: Good call.

PR #379, opened 2018-11-05
[#161465877] hide buy/sell button in the equity portfolio if the acco…

guillaumedebavelaere:
…unt doesn't contain order capabilities

https://www.pivotaltracker.com/story/show/161465877

### DIFF
--- a/TradeItIosTicketSDK2/TradeItPortfolioEquityPositionsTableViewCell.swift
+++ b/TradeItIosTicketSDK2/TradeItPortfolioEquityPositionsTableViewCell.swift
@@ -97,6 +97,12 @@ class TradeItPortfolioEquityPositionsTableViewCell: UITableViewCell {
     // MARK: private
 
     private func updateTradeButtonVisibility() {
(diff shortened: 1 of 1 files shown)


TradeItIosTicketSDK2/TradeItPortfolioEquityPositionsTableViewCell.swift
```diff
@@ -97,6 +97,12 @@ class TradeItPortfolioEquityPositionsTableViewCell: UITableViewCell {
     // MARK: private
 
     private func updateTradeButtonVisibility() {
+        guard let _ = self.selectedPosition?.linkedBrokerAccount.orderCapabilities(forInstrument: TradeItTradeInstrumentType.equities) else {
```
guillaumedebavelaere: There is a `tradeable` flag in the account response we don't have in the `TradeItLinkedBrokerAccount`. Should we add and use this flag instead of checking the capabilities?

guillaumedebavelaere:
@mitochondrion updated

